PR #1, opened 2022-06-08
adding node::Queue; refactoring SinkChannel for code reuse

ryanolson:
In our original design, we had three entities required to form an edge: a Source, Channel and Sink.

This required constructing three object and making two calls to form an edge, pseudo code
```
auto source = s.make_source<T>(...);
auto channel = s.make_channel<T>(...);
auto sink = s.make_sink<T>(...);

s.make_edge(source, channel);
s.make_edge(channel, sink);
```

To reduce complexity for common patterns, it was decided that Sinks would own Channels and provide them to Sources. This had the effect of reducing complexity for edge construction to the creation of a source and a sink and the formation of an edge.

```
auto source = s.make_source<T>(...);
auto sink = s.make_sink<T>(...);

s.make_edge(source, sink);
```

While this reduction in complexity is nice, it also creates a limitation in that each Sink owns its own Channel. This makes load-balancing work across multiple downstream Sinks more difficult, because work is being pushed to each Sink. Pushing has several problems:

- Any write to a Sink's channel which is full, fiber yields the writer.
  - This means if we have N Sinks to which we want to write, we need to have a unique writer per Sink or risk stalling a single writer by blocking on write.
- Sinks operating at different speeds, either from heterogeneous hardware or by the fact the operation being performed may have variable length wall time based on inputs, means push based load-balancing will be problematic.

This MR adds the ability for pull based Sinks using a design similar to the original allowing for standalone `Queue` (Channel holder), but with the default being the current behavior.

```
// Let:
// - SourceProperties<T> be an IngressAcceptor
// - SinkProperties<T> be an IngressProvider

// If:
// - SourceChannel<T> is an IngressAcceptor
// - SinkChannel<T> can be either an IngressProvider or a ChannelAcceptor
// - Queue<T> is an IngressProivder and a ChannelProvider

// Then, Edges can be from:
// - make_edge(IngressAcceptor<T>, IngressProivder<T>)
// - make_edge(ChannelProvider<T>, ChannelAcceptor<T>)
```

This enabled the following example's Sink to be a pull based Sink:

```
        auto source = s.make_object("source", test::nodes::infinite_int_rx_source());
        auto queue  = s.make_object("queue", std::make_unique<node::Queue<int>>());
        auto sink   = s.make_object("sink", test::nodes::int_sink());
        s.make_edge(source, queue);
        s.make_edge(queue, sink);
```

There is no change to the API or default behavior of node/segment objects.

### DIFF
--- a/include/srf/node/sink_channel_base.hpp
+++ b/include/srf/node/sink_channel_base.hpp
@@ -1,0 +1,237 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <srf/channel/buffered_channel.hpp>
+#include <srf/channel/ingress.hpp>
+#include <srf/constants.hpp>
+#include <srf/exceptions/runtime_error.hpp>
+#include <srf/node/edge.hpp>
+#include <srf/node/edge_properties.hpp>
+#include <srf/node/forward.hpp>
+#include <srf/node/sink_properties.hpp>
+#include <srf/utils/type_utils.hpp>
+#include "srf/channel/egress.hpp"
+
+#include <mutex>
+
+namespace srf::node {
+
+/**
+ * @brief Extends SinkProperties to hold a Channel and provide SinkProperties access the the channel ingress
+ *
+ * @tparam T
+ */
+template <typename T>
+class SinkChannelBase
+{
+  protected:
+    SinkChannelBase();
+
+  public:
+    /**
+     * @brief Enables persistence of the Channel.
+     *
+     * The default behavior is to close the owned Channel after all Ingress objects created from channel_ingress have
+     * been destroyed. The persistent option keeps an internal Ingress object live preventing channel closure.
+     *
+     * Invoking stop on the Runner which owns the Runnable explicitly disables persistence.
+     */
+    void enable_persistence();
+
+    /**
+     * @brief Disables persistence of the Channel
+     */
+    void disable_persistence();
+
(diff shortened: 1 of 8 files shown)


include/srf/node/sink_channel_base.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+    // alternative to update_channel
+    // disables ingress_channel
+    void set_shared_channel(std::shared_ptr<channel::Channel<T>> channel);
+
+    // thread safe access to
+    [[nodiscard]] std::shared_ptr<channel::Ingress<T>> ingress_channel();
+
+  private:
+    [[nodiscard]] std::shared_ptr<channel::Ingress<T>> ingress_channel_unsafe();
```
mdemoret-nv, line 108: What is the purpose of `ingress_channel` vs `ingress_channel_unsafe`? Is it just to avoid the deadlock in `enable_persistence`? If so, I feel like a `recursive_mutex` would be a better implementation.

include/srf/segment/resource.hpp
```diff
@@ -0,0 +1,46 @@
...
+    ~Resource() final = default;
+
+  private:
+    ResourceT* get_object() const final
+    {
+        CHECK(m_resource);
+        return m_resource.get();
+    }
+    std::unique_ptr<ResourceT> m_resource;
+};
```
mdemoret-nv, line 44: What is a `Resource`? Why is it needed?

mdemoret-nv (approved):
The design is growing more on me. Just had a couple of comments and the branch needs to be merged with the latest changes.

include/srf/node/sink_channel_base.hpp
ryanolson: Specifically to avoid the recursive mutex.

include/srf/segment/resource.hpp
ryanolson: For lack of a better term, a Resource is not a Runnable. 

segment::Object has pure virtual methods. 

Runnable is the specialization of a segment Object that requires a Runner/Launcher. 

A Resource is constructed, owned and the life cycle is managed by the segment.

include/srf/node/sink_channel_base.hpp
mdemoret-nv: Whats wrong with a recursive mutex? Its not clear at all when you should call `ingress_channel_unsafe` vs `ingress_channel`

include/srf/segment/resource.hpp
mdemoret-nv: Ok, makes sense. But can we choose a different name? `Resource` is such a highly loaded term.

ryanolson: `Component`? `Entity`?

mdemoret-nv: `Component` works.

ryanolson: done

ryanolson:
@gpucibot merge

PR #3, opened 2022-06-08
moved cpp files into cpp folder

ryanolson:


mdemoret-nv (changes requested):
I dont think the extra `srf` in `cpp/srf` is necessary. Most of our include files are then `cpp/srf/include/srf` which is a bit redundant. We should just have
```
cpp/
   include/srf/
   src/
   tests/
```
This is the easiest change for now that gets some clear benefits and follows the RAPIDS layout. We need to have a discussion about organization when we add more libraries in the repo. But for now, this is all we should do.

ryanolson:
If we build multiple .so each with their own set of independent headers, what does that look like?

ryanolson:
The two libraries are a bit intertwined. 

We actually use the templated `node` functionality to build mini internal pipelines.  It's a decoupled pseudo-cyclic dependence. Those internal collections of nodes do not cross the api/abi boundary.

mdemoret-nv:
> Is there an actual use case for this that wouldn't work better than a separate/sub repo?

I think that the goal here would be to separate out certain parts of the library so that you could choose different implementations at runtime. For example, if we had both a `libsrf_core.so` and `libsrf_engine.so` we could have multiple builds of `libsrf_engine.so` with some different features. Like one with compiler optimizations for specific hardware.

mdemoret-nv:
> If we build multiple .so each with their own set of independent headers, what does that look like?

If you had a `core`, `engine` and `architect` libraries, one option would be:

```
cpp/
   include/srf/
      architect/
         architect.hpp
      core/
         core.hpp
      engine/
         engine.hpp
   src/
      architect/
         architect.cpp
         CMakeLists.txt # Includes <srf/architect>. Generates libsrf_architect.so
      core/
         core.cpp
         CMakeLists.txt # Includes <srf/core>. Generates libsrf_core.so
      engine/
         engine.cpp
         CMakeLists.txt # Includes <srf/engine>. Generates libsrf_engine.so
   tests/
      architect/
         CMakeLists.txt # Generates test_srf_architect.so
      core/
         CMakeLists.txt # Generates test_srf_core.so
      engine/
         CMakeLists.txt # Generates test_srf_engine.so
```
Option 2 could be:
```
cpp/
   include/srf/
      architect/
         architect.hpp
      core/
         core.hpp
      engine/
         engine.hpp
   architect/
      src/
         architect.cpp
      tests/
         CMakeLists.txt # Generates test_srf_architect.so
      CMakeLists.txt # Includes <srf/architect>. Generates libsrf_architect.so
   core/
      src/
         core.cpp
      tests/
         CMakeLists.txt # Generates test_srf_core.so
      CMakeLists.txt # Includes <srf/core>. Generates libsrf_core.so
   engine/
      src/
         engine.cpp
      tests/
         CMakeLists.txt # Generates test_srf_engine.so
      CMakeLists.txt # Includes <srf/engine>. Generates libsrf_engine.so
```

I probably favor option 2. The main reason for structuring it like one of these is it keeps all of our installed headers in a single location `#include <srf/...>` and guarantees that there are no header collisions. And at the same time, allows us to build separate libraries.

mdemoret-nv:
Closing since these are too out of date

PR #11, opened 2022-06-08
Set Python3_FIND_STRATEGY and disable the old python policy

### LINKED ISSUE #10: [BUG] Cmake isn't setting Python3_FIND_STRATEGY potentially causing an incorrect Python interpreter to be chosen
Essentially the same as #https://github.com/nv-morpheus/Morpheus/issues/75

dagardner-nv:
fixes #10


mdemoret-nv (approved):
Looks good

PR #18, opened 2022-06-08
Add codable interface for python objects, (Ingress|Egress)Ports python bindings, and other elements required for multi-segment.

drobison00:
Blocked by #75 
- Adds py::object and PyHolder as codable objects
- Support for codable python objects adds some additional utility elements, light wrappers around existing python modules, including multi-process shared memory, and pickle.
- Adds a simple caching mechanism for Python objects on the c++ side which allows for long lived object references, while guaranteeing they're correctly destructed before the python interpreter exits.
- Adds python binding overloads support for `pipeline.make_segment(segment, ["a", "b", "c"], ["x", "y", "z"], init)` -- does not directly support ingress / egress only signatures because we can't differentiate on type and using something like argument count has an ambiguous interpretation.
- Adds IngressPort and EgressPort specializations for py::objects to allow for correct `builder.make_edge` behavior in Python.
- Full set of unit tests for all new objects/code paths.


include/srf/segment/egress_port.hpp
```diff
@@ -61,7 +61,9 @@ class EgressPort final : public Object<node::SinkProperties<T>>,
       m_segment_address(address),
       m_port_name(std::move(name)),
       m_sink(std::make_unique<node::RxNode<T>>())
-    {}
+    {
+        this->m_name = m_port_name;
```
ryanolson, line 65: `ObjectProperties::name()` can be overridden to return `port_name()`

I'm not sure how `segment::Object` a protected member variable `m_name`, but I think that was an oversight.  I think some cpp python proxy object added it instead of deriving from `segment::Object`.

include/srf/segment/ingress_port.hpp
```diff
@@ -56,7 +56,9 @@ class IngressPort : public Object<node::SourceProperties<T>>, public IngressPort
       m_segment_address(address),
       m_port_name(std::move(name)),
       m_source(std::make_unique<node::RxNode<T>>())
-    {}
+    {
+        this->m_name = m_port_name;
```
ryanolson, line 60: See comment for the matching line on EgressPort.

ryanolson (changes requested):
Unfortunately, this requested change is to revert `segment::Object` back to having a pure virtual `name()` method, which was either your change nor mine, but I feel like we need to fix the change that ultimately caused protected member variable to pop up.

include/srf/segment/egress_port.hpp
drobison00: A couple things to note:

1. ObjectProperties::name() is marked final by the Segment::object implementation. Easy enough to remove so that Ingress/Egress can override it, but that doesn't solve all the issues.
2. Runnable sets `this->m_name` in its constructor.
3. A pure virtual setter on ObjectProperties can't be called from within a constructor, so that would mean you'd have to create whatever object, then call obj->set_name("name").

Given the use cases I can see, my current solution, to avoid using protected member variables, is to make `m_name` private, and add a protected `set_name` method to segment::Object. This avoids changing `name()`, eliminates the protected member variable, and allows m_name to be set in the places its used now.

drobison00:
> > Looking into hangs in testing when ingress/egress ports aren't connected.
> 
> I would expect you'd see hangs when they are not connected as the "stop" signal, i.e. the closure of the channel is not allowed to propagate.
> 
> We could probably do more to detect this condition earlier, i.e. incomplete connections, but we need to work under the constraint that Morpheus allows for unconnected Sources as part of Stage construction, i.e. there are no true Sinks.

Double checking that its not specific to py::object ingress/egress types; but, if its not, this does seem like an oversight. I think it makes sense to emit a warning if something isn't connected, but an infinite hang shouldn't be possible.

python/tests/test_edges.py
```diff
@@ -190,5 +190,75 @@ def on_complete():
...
+
+        # This method will get called each time the sink gets a value
+        def sink_on_next(x: MyCustomClass):
+            print("Sink: Got Obj Name: {}, Value: {}".format(x.name, x.value))
+
+        # Build the sink object
+        # sink = seg.make_sink("sink", sink_on_next, None, None)
+        sink = m.SinkBase(seg, "sink")
+
+        seg.make_edge(ingress, sink)
```
mdemoret-nv, line 230: If you replace `MyCustomClass` with a C++ wrapped type, `m.DerivedA`, like you have in the comment, we will get into a situation that I think we want to avoid if we can. Using C++ wrapped types here will cause an unnecessary amount of casing back and forth between python and C++. In a single segment scenario this would go C++ -> C++ and never invoke the GIL. But with multiple segments and ports its going to go C++ -> Python -> Python -> C++ even though we have C++ types and C++ nodes and have the converters registered.

To avoid this, I think we will need some ability to specify the port type from the python side. Something like:
```python
pipeline.make_segment("segment_source", [], [(m.DerivedA, "port1")], segment_source)

pipeline.make_segment("segment_sink", [(m.DerivedA, "port1")], [], segment_sink)
```
where instead of just providing the port name, you provide a tuple of `(port_type, port_name)`. This would allow the runtime the ability to look up the type in pybind11 and see if there was a C++ registered class. If so, we could build the ports in a way that doesnt require the python cast.

python/srf/_pysrf/src/egress_port.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace py = pybind11;
+
+template <>
+EgressPort<pysrf::PyHolder>::EgressPort(SegmentAddress address, PortName name) :
+  m_segment_address(address),
+  m_port_name(std::move(name)),
+  m_sink(std::make_unique<pysrf::PythonNode<pysrf::PyHolder, pysrf::PyHolder>>())
+{
+    this->set_name(m_port_name);
+}
```
mdemoret-nv, line 36: I dont love this template specialization for a couple of reasons:

1. It will be very easy to let the `EgressPort<pysrf::PyHolder>` object get out of sync with the normal `EgressPort<T>` since you have to reimplement the entire class just to change a single line of code.
2. This doesnt allow you to use an internal `PythonNode` with any other type `T` besides `pysrf::PyHolder`

I think this would work better with another template on `EgressPort` that accepts the internal node type to create. Then you can have:

- `EgressPort<PythonNode, pysrf::PyHolder>`
- `EgressPort<RxNode, pysrf::PyHolder>`
- `EgressPort<PythonNode, int>`
- `EgressPort<PythonNode, std::shared_ptr<MultiMessage>>`
- `EgressPort<SnrNode, int>`

python/srf/_pysrf/include/pysrf/egress_port.hpp
```diff
@@ -0,0 +1,8 @@
+//
+// Created by drobison on 6/15/22.
+//
+
+#ifndef SRF_EGRESS_PORT_HPP
+#define SRF_EGRESS_PORT_HPP
+
+#endif  // SRF_EGRESS_PORT_HPP
```
mdemoret-nv, line 8: Did you mean to create this file?

mdemoret-nv (changes requested):
I'm glad to see `make_edge` working, but I am hesitant about the implementation using template specialization. See my comment below, but I think its too fragile of an implementation that will be limiting in performance and flexibility.

I still think modifying the `IngressPorts` and `EgressPorts` object is the way to go. This would provide us an avenue to add more runtime logic from the python side before the `IngressPort`/`EgressPort` is even created.

python/srf/_pysrf/src/ingress_port.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <srf/segment/ingress_port.hpp>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/pytypes.h>
+
+namespace srf::segment {
+namespace py = pybind11;
+
+template <>
```
ryanolson, line 29: I'm struggling to find the difference between this and the definition in the .hpp file.

It feels like this could be replaced with:

```
template class IngressPort<pysrf::PyHolder>;
```

drobison00: The difference is in the RxNode constructor, its built using a PythonNode instead of directly with an RxNode.

python/srf/_pysrf/src/egress_port.cpp
drobison00: > I dont love this template specialization for a couple of reasons:
> 
>     1. It will be very easy to let the `EgressPort<pysrf::PyHolder>` object get out of sync with the normal `EgressPort<T>` since you have to reimplement the entire class just to change a single line of code.
> 
>     2. This doesnt allow you to use an internal `PythonNode` with any other type `T` besides `pysrf::PyHolder`
> 
> 
> I think this would work better with another template on `EgressPort` that accepts the internal node type to create. Then you can have:
> 
>     * `EgressPort<PythonNode, pysrf::PyHolder>`
> 
>     * `EgressPort<RxNode, pysrf::PyHolder>`
> 
>     * `EgressPort<PythonNode, int>`
> 
>     * `EgressPort<PythonNode, std::shared_ptr<MultiMessage>>`
> 
>     * `EgressPort<SnrNode, int>`

Probably worth discussing this in more detail; its not clear to me how this kind of flexibility could work from a Python binding perspective, since all possible combinations would need to be seen at compile time.

python/tests/test_edges.py
drobison00: Resolved by moving adapter's to be runtime constructs.

python/srf/_pysrf/src/segment.cpp
```diff
@@ -184,8 +184,77 @@ std::shared_ptr<srf::segment::ObjectProperties> SegmentProxy::make_sink(srf::seg
...
+    srf::segment::Builder& self,
+    const std::string& name)
+{
+   return self.get_egress<PyHolder>(name);
+}
+
+/*
+std::shared_ptr<srf::segment::ObjectProperties> SegmentProxy::construct_object(srf::segment::Builder& self,
+                                                                        const std::string& name,
+                                                                        std::function<py::object(py::object x)> map_f)
```
dagardner-nv, line 204: Do we need to keep this?

dagardner-nv (approved):
LGTM, had a few random questions.
Thanks for filling out the pydocs

python/srf/_pysrf/include/pysrf/module_wrappers/shared_memory.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include <pybind11/pybind11.h>
+#include <pybind11/pytypes.h>
+
+#include <glog/logging.h>
+
+#include <cstddef>
+#include <string>
+
+using namespace pybind11::literals;
```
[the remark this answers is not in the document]
drobison00: Good catch. I broke a rule I usually look for :+1: .

python/srf/_pysrf/src/segment.cpp
drobison00: I don't think so. Removing it.

python/srf/_pysrf/include/pysrf/module_wrappers/shared_memory.hpp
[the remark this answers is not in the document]
drobison00: Removed these using declarations; they were left over from testing before I separated into hpp/cpp

python/srf/_pysrf/include/pysrf/edge_adapter.hpp
```diff
@@ -0,0 +1,236 @@
...
+ * @tparam SourceT Data type the inheriting source emits
+ */
+template <typename SourceT>
+struct AutoRegSourceAdapter
+{
+    AutoRegSourceAdapter()
+    {
+        // force register_adapter to be called, once, the first time a derived object is constructed.
+        static bool _init = register_adapter();
+    }
```
mdemoret-nv, line 192: This is pretty cool

python/srf/_pysrf/src/utils.cpp
```diff
@@ -221,11 +221,13 @@ pybind11::object&& PyObjectWrapper::move_obj() &&
...
-    {
-        throw srf::exceptions::SrfRuntimeError("Must have the GIL copying to py::object");
-    }
+    // TODO(MDD): Do we need the GIL here? -- (DVN) No; will result in a py::handle to m_obj's m_ptr,
+    //  but the handle only lives as long as m_obj has a non-zero ref count. Does seem dangerous to do
+    //  this via operator though, seems like somewhat unexpected behavior.
+    // if (PyGILState_Check() == 0)
+    // {
+    //     throw srf::exceptions::SrfRuntimeError("Must have the GIL copying to py::object");
+    // }
```
mdemoret-nv, line 230: If you identified that this isnt necessary, you can remove this block.

mdemoret-nv (changes requested):
A couple of really small changes

python/tests/test_node.py
```diff
@@ -76,7 +76,8 @@ def node_fn(x: int):
     options = srf.Options()
 
     # Set to 1 thread
-    options.topology.user_cpuset = "0-{}".format(pe_count)
+    # TODO(#38)
+    # options.topology.user_cpuset = "0-{}".format(pe_count)
```
mdemoret-nv, line 80: I cannot repo this locally and these changes are crucial to the test functioning correctly. This also has been passing in our CI. Do you have more info on what is causing this? I would prefer to keep these enabled if they are working in CI

drobison00: Will double check; it may have been fixed at some point with another patch.

python/tests/test_edges.py
```diff
@@ -190,5 +192,79 @@ def on_complete():
...
+    # Register pipeline to tell executor what to run
+    executor.register_pipeline(pipeline)
+
+    # This will start the pipeline and return immediately
+    executor.start()
+
+    # Wait for the pipeline to exit on its own
+    executor.join()
+
+    print("Pipeline exited!")
```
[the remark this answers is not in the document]
drobison00: Done

python/tests/test_edges.py
```diff
@@ -22,9 +24,11 @@ def segment_init(seg: srf.Builder):
         source = m.SourceDerivedB(seg, "source")
 
         node = m.NodeBase(seg, "node")
+        print("Makeing edge between source and node")
         seg.make_edge(source, node)
 
         sink = m.SinkBase(seg, "sink")
+        print("Makeing edge between node and sink")
```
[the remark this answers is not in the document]
drobison00: Done

python/tests/test_edges.py
```diff
@@ -190,5 +192,79 @@ def on_complete():
...
+    # Create a segment
+    pipeline.make_segment("segment_source", [], ["port1"], segment_source)
+
+    pipeline.make_segment("segment_sink", ["port1"], [], segment_sink)
+
+    # Build executor options
+    options = srf.Options()
+
+    # Set to 1 thread
+    options.topology.user_cpuset = "0-1"
```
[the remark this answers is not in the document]
drobison00: Fixed

include/srf/node/edge_builder.hpp
```diff
@@ -18,21 +18,90 @@
...
+     * uses it.
+     * @param source
+     * @param sink
+     * @param ingress_handle Ingress handle for the sink object
+     * @return Ingress handle constructed by the adapter
+     */
+    static std::shared_ptr<channel::IngressHandle> default_ingress_adapter_for_sink(
+        srf::node::SourcePropertiesBase& source,
+        srf::node::SinkPropertiesBase& sink,
+        std::shared_ptr<channel::IngressHandle> ingress_handle);
```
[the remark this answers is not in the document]
drobison00: Makes sense; although if we do that I'll probably just delete the defaults entirely.

python/srf/_pysrf/include/pysrf/edge_adapter.hpp
ryanolson: Is _varname the form we use for static members?

python/tests/test_node.py
drobison00: This looks like it was fixed somewhere else. Will close the other bug.

mdemoret-nv:
@gpucibot merge

PR #21, opened 2022-06-13
Add Support for RAPIDS 21.10

mdemoret-nv:
This PR adds support to the build matrix for RAPIDS `21.10`, `22.04`, and `22.06`. In the future only the matching version + the previous RAPIDS release will be supported (i.e. for the `22.08` release, we will only support `22.08` and `22.06`).

The main problem child here is `cudf=22.06` which upgrades to `arrow-cpp=7`. The `arrow-cpp=7` package does not follow the standard library version for `conda_build_config.yaml` and requires `grpc-cpp=1.45` and `glog=0.6`.


ryanolson (approved):

drobison00 (approved):
Looks fine. Maybe make the libprotobuf version explicit?

mdemoret-nv:
> Looks fine. Maybe make the libprotobuf version explicit?

We dont want to set the version of libprotobuf because we want whatever version is required by `grpc-cpp`.

PR #25, opened 2022-06-13
Fix Pins for abseil-cpp, grpc-cpp, and glog

mdemoret-nv:
Another package must have been released between the previous MR #21 and trying to release a new conda package. This changed which dependencies were used in the build. This PR explicitly sets the package versions for all the necessary packages to prevent this issue.

Note: This requires that we use the default versions of abseil-cpp, grpc-cpp, and glog for cuDF.


dagardner-nv (approved):
LGTM

PR #28, opened 2022-06-14
adding PortGraph; validate Pipeline definitions

### LINKED ISSUE #27: [BUG] Segments created with unconnected ingress ports hang on shutdown.
**Describe the bug**
Creating a segment that has unconnected IngressPorts will cause an infinite hang at segment shutdown.

**Steps/Code to reproduce bug**
gtest
```c++
TEST_F(TestPipeline, TestThing) {
    std::function<void(srf::segment::Builder&)> init = [](srf::segment::Builder& builder) {};

    pysrf::Pipeline pipe;

    pipe.make_segment("TestSegment1", {"some_ingress"}, {}, init);

    auto opt1 = std::make_shared<srf::Options>();
    opt1->topology().user_cpuset("0");
    opt1->topology().restrict_gpus(true);

    srf::Executor exec1{opt1};

    exec1.register_pipeline(pipe.swap());

    exec1.start();
    exec1.join();
}
```
```bash
gtest ./test_script.x

... snip ...
I20220613 17:03:05.619365 2935157 instance.cpp:221] [Segment: 38357/0/2513764352] join complete
I20220613 17:03:05.619374 2935157 controller.cpp:144] pipeline::controller: update complete
I20220613 17:03:05.619376 2935151 instance.cpp:226] [Segment: 38357/0/2513764352] join started
I20220613 17:03:05.619382 2935151 instance.cpp:244] [Segment: 38357/0/2513764352] awaiting on ingress port join to some_ingress

... hangs here ...
```

**Expected behavior**
Should either emit a message indicating that it is an error to create unconnected ingress nodes and either execute to completion, cleaning up correctly, or exit immediately.

**Environment overview (please complete the following information)**
Conda environment with debug SRF build.

ryanolson:
Fixes #27 by throwing an exception when registering a pipeline with an incomplete or currently invalid port configuration.

### DIFF
--- a/src/internal/executor/executor.cpp
+++ b/src/internal/executor/executor.cpp
@@ -19,10 +19,13 @@
 
 #include "internal/pipeline/manager.hpp"
 #include "internal/pipeline/pipeline.hpp"
+#include "internal/pipeline/port_graph.hpp"
 #include "internal/pipeline/types.hpp"
 #include "internal/resources/resource_partitions.hpp"
 #include "internal/system/system.hpp"
+#include "internal/utils/contains.hpp"
 #include "srf/core/addresses.hpp"
+#include "srf/exceptions/runtime_error.hpp"
 #include "srf/internal/pipeline/ipipeline.hpp"
 #include "srf/options/options.hpp"
 #include "srf/types.hpp"
@@ -32,6 +35,8 @@
 #include <map>
 
 namespace srf::internal::executor {
+
+static bool valid_pipeline(const pipeline::Pipeline& pipeline);
 
 Executor::Executor(Handle<Options> options) :
   m_system(system::make_system(std::move(options))),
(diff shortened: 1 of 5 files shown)


ryanolson:
ok to test

src/internal/executor/executor.cpp
```diff
@@ -91,4 +102,41 @@ void Executor::do_service_await_join()
...
+        {
+            valid = false;
+            // todo - print list of segments names for ingress/egres connections to this port
+            LOG(WARNING) << "port: " << name << " has incomplete connections - used as ingress on "
+                         << connections.ingress_segments.size() << " segments; used as egress on "
+                         << connections.egress_segments.size() << " segments";
+        }
+
+        // we currently only have an load-balancer manifold
+        // it doesn't make sense to connect segments of different types to a load-balancer, they should probably be
```
drobison00, line 126: What 'types' of segments are there? Is this just referencing the type of the ingress/egress port on the segment, or something else?

drobison00 (approved):
LGTM.

src/internal/executor/executor.cpp
ryanolson: By types, I mean two different segments with the same port name.

https://github.com/ryanolson/srf/blob/validate_ports/src/tests/test_pipeline.cpp#L324-L325

When you have two different types of segments, it's unclear that load-balancing between them makes sense. We definitely need more information to handle this condition properly.

ryanolson:
@gpucibot merge

PR #40, opened 2022-06-17
pimpl IBuilder

ryanolson:
replacing the virtual interface of IBuilder with a pimpl interface to improve abi compatibility


cwharris:
I don't understand. This adds another level of indirection, but why? What problem is it solving that simple virtual class inheritance and a `std::shared_ptr` won't?

ryanolson:
Additions to the pure virtual interface would break the abi.

These I-classes in `srf/internal` are the interfaces between the frontend and the backend. Probably poorly named. That which is in internal will likely be moved to it's own separate library.

mdemoret-nv:
Before we merge any PRs related to stabilizing the ABI, I think we need to come up with guidelines potentially rename/relocate some of the `I` classes. It's also not very clear to me either why it's organized the way that it is. The inheritance and implementation classes seem to be all over the place

mdemoret-nv:
> I agree we need to organize and rename them, but that shouldn't block this update. Otherwise it will just be code rot.

The problem with this update is it's not consistent with the other PIMPL interfaces:

- `srf::pipeline::Pipeline final : public internal::pipeline::IPipeline`
- `srf::segment::Definition final : public internal::segment::IDefinition`
- `srf::executor::Executor final : public internal::executor::IExecutor`
- `srf::segment::Builder`
   - This used to inherit from `internal::segment::IBuilder` which was consistent

So which way should we be doing it? Without guidelines this is hard to say what it should be.

Looking at this PR in more detail, `Builder` seems to be different because it accepts the `IBuilder backend` in the constructor and combines the backend implementation with the PIMPL interface, which is very confusing. Especially since both the front end interface and IMPL classes are called `Builder`.

I think things would be a lot clearer if the PIMPL object was separated from the backend object so the pattern used for `IBuilder` was the same as `IPipeline`, `IDefinition`, etc. Maybe a `IBuilderBackend` object?

ryanolson:
The other are all pimpl'd with a single pure virtual destructor to forbid construction without inheritance.

include/srf/internal/segment/ibuilder.hpp
```diff
@@ -23,6 +23,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include "srf/utils/macros.hpp"
```
drobison00, line 26: Should go above standard includes.

drobison00 (commented):
I'd actually missed that the internal and public classes have the same name.

This gets really confusing, and I'd like to take a look at naming conventions, or isolation conventions between public and detail interfaces.

mdemoret-nv:
@gpucibot merge

PR #42, opened 2022-06-19
adding Resuable<T> and ReusablePool<T>

ryanolson:
This PR add two new classes to `srf::data::`
- `Reusable<T>`
- `SharedReusable<T>`
- `ReusablePool<T>`

`Reusable<T>` is a move only holder of a `unique_ptr<T>` which can be acquired from `ReusablePool<T>::await_item()`.  The held `unique_ptr<T>` is returned to the pool when `Reusable<T>` is destroyed while still holding a `unique_ptr<T>`.

`Reusable<T>` can be converted to `SharedReusable<T>` which is immutable, but also copyable and reference counted. The destruction of a `SharedReusable<T>` returns the object `T` to the `ReusablePool`.

These concepts can be furthered extended with C++20 `concepts`


src/tests/nodes/common_nodes.hpp
```diff
@@ -17,15 +17,74 @@
...
+    {
+        return m_buffer.size();
+    }
+
+  private:
+    std::array<char, Bytes> m_buffer;
+};
+
+template <std::size_t Bytes>
+class BufferSource final : public srf::node::GenericSource<srf::data::Reusable<Buffer<Bytes>>>
```
drobison00, line 58: Somewhat of a tangent, but I wasn't aware we still had GenericNodes. Unless there is some kind of functionality that can't be achieved using an RxNode, I'm kind of opposed to creating/using multiple core node types.

src/tests/test_pipeline.cpp
```diff
@@ -356,3 +357,100 @@ TEST_F(TestPipeline, RequiresMoreManifolds)
...
+        return m_buffer.size();
+    }
+
+  private:
+    std::array<std::size_t, 1024> m_buffer;
+};
+
+TEST_F(TestPipeline, ReusablePool)
+{
+    auto pool = data::ReusablePool<Buffer>::create(32);
```
drobison00, line 389: Would like to see additional unit tests, at least one for each new object, and which exercise the full API so its clear its working as expected.

drobison00 (changes requested):
A couple questions, I'd like to split ReusablePool objects into a new unit test file, and needs additional unit tests to cover the new objects more fully.

ryanolson:
Yes, probably true.

include/srf/data/reusable_pool.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+#pragma once
+
+#include "srf/channel/buffered_channel.hpp"
+#include "srf/utils/macros.hpp"
+
+#include <glog/logging.h>
+
+#include <atomic>
+#include <boost/fiber/buffered_channel.hpp>
```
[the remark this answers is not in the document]
ryanolson: Nice catch. There are defects filters that format can apply to help with this I think.

include/srf/data/reusable_pool.hpp
```diff
@@ -0,0 +1,243 @@
...
+            m_on_return_fn(*item);
+        }
+        m_channel.push(std::move(item));
+    }
+
+    std::mutex m_mutex;
+    std::size_t m_size;
+    const std::size_t m_capacity;
+    std::function<void(T&)> m_on_return_fn{nullptr};
+    boost::fibers::buffered_channel<item_t> m_channel;
```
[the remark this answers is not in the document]
ryanolson: buffered_channel is highly optimized and naturally provides the yielding behavior we want.

include/srf/data/reusable_pool.hpp
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * @brief Number of items managed by the pool
+     */
+    std::size_t size() const
+    {
+        return m_size;
+    }
+
+  private:
+    ReusablePool(std::size_t capacity, on_return_t on_return_fn) :
```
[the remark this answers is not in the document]
ryanolson: It inherits from enable_shared_from_this which requires the object to be held by a shared ptr. 

The static create method enforces this behavior.

include/srf/data/reusable_pool.hpp
```diff
@@ -0,0 +1,243 @@
...
+    friend Reusable<T>;
+    friend SharedReusable<T>;
+};
+
+template <typename T>
+class Reusable final
+{
+    using pool_t = ReusablePool<T>;
+
+    Reusable(std::unique_ptr<T> data, std::shared_ptr<pool_t> pool) : m_data(std::move(data)), m_pool(std::move(pool))
```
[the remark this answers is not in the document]
ryanolson: You can only acquire a Resusable by acquiring it from the pool. 

This helps to enforce the limit on the number of objects managed by a pool. 

The pool does not directly track the umber of objects "checked out".

src/tests/nodes/common_nodes.hpp
drobison00: Is there some specific need to support conventional inheritance for RxNodes? This seems like something a 3rd party consumer could do if they really wanted to; but kind of redundant/confusing in the base library.

ryanolson: It's a convenience class for those that do not know or want to know rx.

ryanolson: Off-topic - Marking as resolved.

include/srf/data/reusable_pool.hpp
```diff
@@ -0,0 +1,250 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <srf/channel/buffered_channel.hpp>
+#include <srf/utils/macros.hpp>
+
+#include <glog/logging.h>
```
drobison00, line 23: Includes need to be sorted.

src/tests/test_next.cpp
```diff
@@ -29,9 +29,12 @@
 #include "srf/segment/object.hpp"
 #include "srf/utils/macros.hpp"
 
+#include <iterator>
```
drobison00, line 32: Needs to go at the bottom with other std includes

src/tests/test_next.cpp
```diff
@@ -29,9 +29,12 @@
...
 #include <srf/channel/egress.hpp>
 #include <srf/channel/ingress.hpp>
 #include <srf/channel/status.hpp>
+
```
drobison00, line 36:
```suggestion
```

drobison00 (commented):
Once headers are fixed, should be good to go.

ryanolson:
@drobison00 - i'm holding off on any header changes in this PR until #47 is finished. 

When I merge from main, the new clang-format rules should fix the include blocks.

ryanolson:
merged 22.08 with formatting from #47.  this should have cleaned up all the headers.

drobison00 (approved):

ryanolson:
@gpucibot merge

PR #43, opened 2022-06-19
Allow service.cpp reuse forward_state more.

cwharris:


ryanolson:
More additions to the correct behavior of this state machine
- if stop/kill is called before start, service should be marked as complete and calls to start should be a no-op, warning or error.

If we only allows a single await join to be called, then we could do:
- Service could hold a Status object and possibly a exception_ptr.  
- The status object OK or CANCELLED could be returned on await_join unless a exception_ptr was valid
- If an exception_ptr was held, then rethow that on await_join.

If await join could have multiple callers, then we have to have an ERROR on the status and a means for only one of the awaiters to get the exception_ptr.

cwharris:
I think we may have multiple ideas conflated here. One idea is the internal state of the service (whether it has been started, if it is in the process of stopping, or perhaps being killed), and the other is the idea of an external mechanism depending on the completion of the service (be it gracefully through natural completion, gracefully through stop, or ungracefully through kill). I wonder if it would be cleaner to remove the Awaiting state entirely, and semantically have any await trigger a stop. Awaits could also be provided with a value indicating the severity of the stop (natural/stop/kill).

cwharris:
> If await join could have multiple callers, then we have to have an ERROR on the status and a means for only one of the awaiters to get the exception_ptr.

I don't think semantically it makes sense to force a single caller to handle the exception. We could instead provide the exception to all awaiters, and allow the callers to determine which awaiter should handle the exception. The callers will know more about the situation than the service will. What if 3 callers await, 2 of which do not care if there is an exception, and 1 of which does care? How do we ensure that call is the one to handle the exception?

PR #44, opened 2022-06-19
Remove unnecessary CHECKs and friend-ship from IExecutor

cwharris:


src/internal/executor/iexecutor.cpp
```diff
@@ -30,32 +30,27 @@ namespace srf::internal::executor {
 IExecutor::IExecutor(std::shared_ptr<Options> options) : m_impl(make_executor(std::move(options))) {}
 IExecutor::IExecutor(std::unique_ptr<system::ISystem> system) : m_impl(make_executor(system::System::unwrap(*system)))
 {
-    CHECK(m_impl);
```
ryanolson, line 33: This check should probably stay, since it would be an error condition if `make_executor` returned a `nullptr`.

ryanolson: A default constructor should probably be used which takes a `unique_ptr<Executor>`, then the check for not nullptr only needs to be on the default constructor.

include/srf/internal/executor/iexecutor.hpp
```diff
@@ -50,7 +49,6 @@ class IExecutor
 
   private:
     std::shared_ptr<Executor> m_impl;
```
ryanolson, line 51: Should probably be `unique_ptr`

ryanolson:
Thanks for cleaning up the class.  See the comments above.

src/internal/executor/iexecutor.cpp
```diff
@@ -27,35 +27,34 @@
...
 
 void IExecutor::register_pipeline(std::unique_ptr<internal::pipeline::IPipeline> pipeline)
 {
-    CHECK(m_impl);
     m_impl->register_pipeline(std::move(pipeline));
 }
 
 void IExecutor::start()
 {
-    CHECK(m_impl);
```
ryanolson, line 46: I think these need to stay.

```
srf::Executor executor(options)
parent->set_executor(std::move(executor));
executor.start(); // segfaults without checks; checks provide clearer reason for failure
```

cwharris: This seems more like a code smell than desired behavior. When and why would we want to swap out the executor, and why wouldn't set_executor be responsible for the `CHECK`?

ryanolson: This is just move semantics. 

I don't think we can mark Executor as not-moveable.

cwharris: It appears we can, and probably should? I don't see why something as high level as Executor would need to be moved when we're heap allocating it and keeping it in a `unique_ptr`.

cwharris: it's a `unique_ptr`, It's privately owned, and we're not moving it. Why are we checking that it's not null every time we use it? We know it's not null because we're not nullifying it, we're not moving it, and no-one else has access to it. I'm confused why this check is necessary.

src/internal/executor/iexecutor.cpp
mdemoret-nv: FYI, you can use ` m_impl(CHECK_NOTNULL(make_executor(system::System::unwrap(*system))))` to perform the check inline

src/internal/executor/iexecutor.cpp
ryanolson: it's owned by a class that is moveable both by construction and assignment.

```
auto exec2 = std::move(exec1)
```

`exec2` now owns the unique_ptr held by `exec1`; `exec1`'s unique_ptr is null.

cwharris: We discussed offline, and it sounds like we're good to go with these changes, assuming Executor can have it's move-ability and copy-ability deleted, at least until we find a use case for move-ability of Executors.

PR #50, opened 2022-06-21
Add a benchmark stage to CI

### LINKED ISSUE #37: [FEA] Add benchmarks to CI
We should run the benchmarks in a stage parallel to the test stage.

Depends on #4

**Describe the solution you'd like**
Ideally we would setup the conda env in a separate step, such that we could infer the runtime of the benchmarks from the runtime of the step.

Benchmarks could be located with:
```bash
BENCHMARKS=($(find ${SRF_ROOT}/build/benchmarks -name "*.x"))
```

@drobison00 suggested pinning the benchmarks to a single core via `numactl --physcpubind=0 <bench cmd>`

We want to use the json output via the `--benchmark_out_format=json` flag

dagardner-nv:
Executes benchmarks in a stage parallel to the test stage.
The benchmark stage is separated into three steps:
- pre_benchmark.sh - downloads build from previous stage
- benchmark.sh - executes benchmarks
- post_benchmark.sh - archives benchmark reports to S3 as `benchmark_reports.tar.bz`

The idea is as much as possible the runtime of the step should represent the runtime of the benchmarks.

https://downloads.rapids.ai/ci/srf/pull-request/50/ce8e07f00b002a8c2a97dd707ec3a6181f6e3c20/x86_64/gcc/benchmark_reports.tar.bz

Closes #37


ryanolson:
Are we just benchmarking on the CI machine for the sake of completeness to ensure the benchmark code is not broken?

If so, great.

But we should not use the results of the benchmarks run by the CI as a performance base line.

drobison00 (approved):
Looks good.

ci/scripts/jenkins/benchmark.sh
```diff
@@ -0,0 +1,45 @@
...
+       gpuci_logger "${bench_name} completed with an exit status of ${BENCH_RESULT}"
+       BENCH_RESULTS=$(($BENCH_RESULTS+$BENCH_RESULT))
+
+       set -e
+done
+
+# We want the archive step to run, even if we failed, save our exit status, which the next step can use
+echo ${BENCH_RESULTS} > ${WORKSPACE_TMP}/exit_status
+
+gpuci_logger "Done"
```
mdemoret-nv, line 45:
```suggestion
gpuci_logger "Running Benchmarks... Done"
```
I like to keep the start and end logging messages like this in pairs so its easy to search for and clear what "Done" is referring to.

Then you see stuff in the log like:
```
Running Benchmarks...
...
Running Benchmarks... Done
```

ci/scripts/jenkins/post_benchmark.sh
```diff
@@ -0,0 +1,30 @@
...
+REPORTS_DIR="${WORKSPACE_TMP}/reports"
+
+gpuci_logger "Archiving benchmark reports"
+cd $(dirname ${REPORTS_DIR})
+tar cfj ${WORKSPACE_TMP}/benchmark_reports.tar.bz $(basename ${REPORTS_DIR})
+
+gpuci_logger "Pushing results to ${DISPLAY_ARTIFACT_URL}/"
+aws s3 cp ${WORKSPACE_TMP}/benchmark_reports.tar.bz "${ARTIFACT_URL}/benchmark_reports.tar.bz"
+
+exit $(cat ${WORKSPACE_TMP}/exit_status)
```
mdemoret-nv, line 30: This seems really odd to me. Is there no other way to pass information from one script to the next? Could we not have a `benchmark_wrapper.sh` script that does:
```bash
# Do pre-benchmark stuff here

# Run benchmark
${WORKSPACE}/ci/scripts/jenkins/benchmark.sh
BENCHMARK_EXIT_STATUS=$?

# Post benchmark stuff
```

mdemoret-nv (changes requested):
The pre/post benchmark stuff seems odd to me and out of place. Couple of quick questions and one suggestion.

ci/scripts/jenkins/post_benchmark.sh
dagardner-nv: I don't think so not with the way its setup currently with the three scripts as separate steps.

ryanolson:
> Mostly to ensure that they aren't broken, but while we can't use the performance as a precise baseline, it would be good to know that they run and complete rather than deadlocking. 

Certainly it's better than nothing. We could use the archived benchmark data to help determine where/when a regression occurred with the understanding that the variability run-to-run maybe higher than if run on a dedicated system. 

It will be interesting to see the magnitude of the run-to-run variance.

mdemoret-nv (approved):

PR #73, opened 2022-06-27
Changing RAPIDS_VERSION to SRF_RAPIDS_VERSION

mdemoret-nv:
The latest Conda packages upgraded to RAPIDS CMake 22.06 which has a bug when trying to install UCX. It uses the following for the export set:

```Cmake
find_package(ucx 1.12 QUIET)
find_dependency(ucx)
```

The `ucx` installed package doesnt check for existing targets so the second call to `find_dependency` fails with an error. This explicitly sets: `rapids_export_package` which will change the export set to:

```Cmake
find_dependency(ucx)
```

This also resolves many hidden issues by renaming the variable `RAPIDS_VERSION` to `SRF_RAPIDS_VERSION`.


ryanolson (approved):

mdemoret-nv:
@gpucibot merge

PR #83, opened 2022-06-28
Fix for pipelines beginning before start is called

### LINKED ISSUE #80: [BUG] Pipelines starts before Executor::start is called
what is happening:

`internal::pipeline::Manager` starts the service in the constructor, which means the `Controller` live when an update is received on registering the pipeline.

what should happen:

- pipeline::Manager is created on pipeline registration.
- the construct can construct the channel so updates can be pushed
- edge formation and controller creation should delayed until `do_service_start`

ryanolson:
Fixes #80


mdemoret-nv (approved):
Looks good. Would like a test to enforce this in the future. I have created issue #84 to add that at a later date.

mdemoret-nv:
@gpucibot merge

PR #107, opened 2022-06-30
Transient Pool

ryanolson:
```cpp
/**
 * @brief ReusablePool of memory::buffers that are used as reusable reference-counted monotonic memory resources
 *
 * TransientPool is a ReusablePool of memory::buffers from which smaller buffers are allocated similar to a monotonic
 * memory resource, i.e. pointer pusing stack; however the TransientBuffer or Transient<T> object pulled from the pool
 * hold a SharedReusable<memory::buffer> which keeps the entire monotonic stack from returning to the ReusablePool
 * until all transient objects created on a given stack are deallocated.
 *
 * Allocation of Transient object should be incredibly fast; even faster than the Reusable/SharedResuable on which they
 * are based, since a single Reusable<memory::buffer> might back 10s-1000s of allocations depending on size.
 *
 * It is critical that all Transient object allocated from a pool have similar life cycles
 */
```

### DIFF
--- a/src/internal/memory/transient_pool.hpp
+++ b/src/internal/memory/transient_pool.hpp
@@ -1,0 +1,142 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "srf/data/reusable_pool.hpp"
+#include "srf/memory/buffer.hpp"
+#include "srf/memory/resources/memory_resource.hpp"
+#include "srf/utils/macros.hpp"
+
+#include <glog/logging.h>
+
+#include <cstddef>
+#include <new>
+#include <utility>
+
+namespace srf::internal::memory {
+
+/**
+ * @brief A short-lived buffer based on a portion of a data::SharedResable<memory::buffer>
+ *
+ * @see TransientPool for more details.
+ */
+
+class TransientBuffer
+{
+  public:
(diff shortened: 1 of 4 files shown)


src/internal/memory/transient_pool.hpp
```diff
@@ -0,0 +1,148 @@
...
+{
+  public:
+    TransientBuffer(void* addr, std::size_t bytes, srf::data::SharedReusable<srf::memory::buffer> buffer);
+    ~TransientBuffer();
+
+    DELETE_COPYABILITY(TransientBuffer);
+    DEFAULT_MOVEABILITY(TransientBuffer);
+
+    void* data();
+    std::size_t bytes() const;
```
dagardner-nv, line 49: Maybe just a comment of what bytes represents here, is it the size of `data` or is it the total size of `TransientBuffer`>

Also is `addr` null terminated?

src/internal/memory/transient_pool.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+  private:
+    T* m_data;
+};
+
+/**
+ * @brief ReusablePool of memory::buffers that are used as reusable reference-counted monotonic memory resources
+ *
+ * TransientPool is a ReusablePool of memory::buffers from which smaller buffers are allocated similar to a monotonic
+ * memory resource, i.e. pointer pusing stack; however the TransientBuffer or Transient<T> object pull from the pool
```
dagardner-nv, line 114:
```suggestion
 * memory resource, i.e. pointer pushing stack; however the TransientBuffer or Transient<T> object pull from the pool
```

dagardner-nv (changes requested):
Looks good, but can you add a few more comments to the header?

src/internal/memory/transient_pool.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * @see TransientPool and TransientBuffer for more details.
+ */
+
+template <typename T>
+class Transient final : private TransientBuffer
+{
+  public:
+    Transient(TransientBuffer&& buffer) : TransientBuffer(std::move(buffer))
+    {
+        CHECK_LT(sizeof(T), bytes());
```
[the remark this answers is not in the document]
ryanolson: the buffer should be allocated to be a bit larger than sizeof(T) to allow for proper alignment.

it should be

CHECK_LE(sizeof(T) + alignof(T), bytes());

ryanolson, line 71:
```suggestion
        CHECK_LE(sizeof(T) + alignof(T), bytes());
```

dagardner-nv (approved):

ryanolson:
@dagardner-nv - i added explicit move constructors and assignment operators since the default ones do not play nice with raw pointers.

I also added the ability to pass arguments for the `await_object<T>(...)` call.

Let me know if you think of any thing I might have missed.

PR #110, opened 2022-07-02
Transient Pool

ryanolson:
```cpp
/**
 * @brief ReusablePool of memory::buffers that are used as reusable reference-counted monotonic memory resources
 *
 * TransientPool is a ReusablePool of memory::buffers from which smaller buffers are allocated similar to a monotonic
 * memory resource, i.e. pointer pusing stack; however the TransientBuffer or Transient<T> object pulled from the pool
 * hold a SharedReusable<memory::buffer> which keeps the entire monotonic stack from returning to the ReusablePool
 * until all transient objects created on a given stack are deallocated.
 *
 * Allocation of Transient object should be incredibly fast; even faster than the Reusable/SharedResuable on which they
 * are based, since a single Reusable<memory::buffer> might back 10s-1000s of allocations depending on size.
 *
 * It is critical that all Transient object allocated from a pool have similar life cycles
 */
```


ryanolson:
This was already approved as #107 but was closed by the CI bot.

ryanolson:
@gpucibot merge

mdemoret-nv (approved):

mdemoret-nv:
@gpucibot merge

PR #113, opened 2022-07-04
Internal Runtime Query + CPP Checks

ryanolson:
Enables internal members to query the runtime resources from a `static thread_local` member of the `resources::Manager`

This builds on #112

This diff will highlight what's in this PR:
https://github.com/nv-morpheus/SRF/compare/pull-request/112...pull-request/113

Appended to this PR is a repo wide update to correct all cpp_checks that are now failing.


src/internal/resources/manager.cpp
```diff
@@ -134,6 +148,18 @@ Manager::Manager(std::unique_ptr<system::Resources> resources) :
...
     }
+
+    // set thread local access to resources on all fiber task queues and any future thread created by the runtime
+    for (auto& partition : m_partitions)
+    {
+        m_system->register_thread_local_initializer(partition.partition().host().cpu_set(), [this, &partition] {
+            m_thread_resources = this;
+            if (system().partitions().device_to_host_strategy() == PlacementResources::Dedicated)
+            {
+                m_thread_partition = &partition;
```
drobison00, line 159: What is the significance of this; what would we observe if PlacementResources is something other than dedicated

src/internal/resources/manager.cpp
```diff
@@ -151,4 +177,38 @@ PartitionResources& Manager::partition(std::size_t partition_id)
...
+        {
+            auto& resources = Manager::get_resources();
+
+            if (resources.system().partitions().device_to_host_strategy() == PlacementResources::Shared)
+            {
+                LOG(ERROR) << "runtime partition query is disabed when PlacementResources::Shared is in use";
+                throw exceptions::SrfRuntimeError("unable to access runtime parition info");
+            }
+
+            LOG(ERROR) << "thread with id " << std::this_thread::get_id()
```
drobison00, line 206: If this is something that we really want to prevent; would it be worthwhile to store the thread id on the resource object, and compare it with the requesting thread, rather than this kind of implicit check?

drobison00 (approved):
LGTM Once #112 goes in.

src/internal/resources/manager.cpp
ryanolson: This is the primary concern of #114. 

My vote to to remove shared and dedicated and keep only dedicated functionality. 

The edge cases where you have more GPUs than cores we could either allow oversubscription or limit the number of devices include in the runtime resources.

drobison00 (approved):
LGTM

ryanolson:
@gpucibot merge

PR #115, opened 2022-07-04
Codable Resources

ryanolson:
This PR bring together the internal runtime resources with the codable protocol.

Object being serialized via the codable protocol require the runtime resources for:
- determining if a memory block is registered with the ucx context
- allocating a registered host/device buffer

This breaks the old codable tests which were run on the main thread. The main thread is not part of the SRF runtime. The fix to this issue is to ensure all serialization/deserialization occurs on SRF runtime threads. To fix the tests, simply perform the action on the main thread queue or create a custom runnable.

Lists below is the change to any interfaces required for better error handling:
- [x] registration cache - return an optional ucx::MemoryBlock instead of a const ref to a block or an exception


drobison00 (commented):
Looks fine overall, once #113 goes in and the merge conflicts are resolved I want to double check coverage and I'll approve.

ryanolson:
This PR is blocked by the tests in https://github.com/ryanolson/srf/blob/codable_resources/python/srf/_pysrf/tests/test_codable_pyobject.cpp

The breaking change to the codable protocol is that we need a SRF runtime to be instantiated so the codable methods have access to network registered memory resources (allocators) and the ucx registration cache.

drobison00 (commented):
Looks fine overall; could you elaborate a bit on what needs to be propagated to the pysrf tests to make them work?

PR #117, opened 2022-07-05
Fix `flatten()` Operator With Infinite Sources

### LINKED ISSUE #116: [BUG] `flatten()` Operator Does Not Push Messages Until Upstream Has Completed
**Describe the bug**
The python `flatten()` operator was changed to use `concat_map()` and `iterate()` instead of building our own operator in a recent release. This works in all unit tests because we use finite source nodes. If you change any of these to use an infinite source, then no messages will get passed on by the operator until the source completes.

It's unclear whether its the `concat_map`, `iterate` or some other problem causing the issue. Reverting back to the old style of `flatten()` operator fixes the issue.

**Steps/Code to reproduce bug**
Creating an infinite source of arrays and then a node with `flatten` operator in python will illustrate the issue. Reproducing this issue can be difficult since there is no way to remove the buffers between nodes in python. 

**Expected behavior**
`flatten()` should push the flattened messages before `on_completed()` is called.

mdemoret-nv:
This is a hotfix to close #116. This PR reverts the implementation back to the one that does not use `concat_map` and `iterate`. This also ensures all `on_error` and `on_completed` are forwarded.

### DIFF
--- a/python/srf/_pysrf/src/operators.cpp
+++ b/python/srf/_pysrf/src/operators.cpp
@@ -65,27 +65,52 @@ PythonOperator OperatorsProxy::flatten()
 {
     //  Build and return the map operator
     return PythonOperator("flatten", [=](PyObjectObservable source) {
-        return source.concat_map([=](PyHolder data_object) {
-            AcquireGIL gil;
-
-            // Convert to a vector to allow releasing the GIL
-            std::vector<PyHolder> obj_list;
-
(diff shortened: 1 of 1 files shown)


mdemoret-nv:
Merging without review due to Morpheus 22.06 release deadline.

mdemoret-nv:
@gpucibot merge

PR #122, opened 2022-07-05
Remote Descriptor Manager

ryanolson:


src/internal/remote_descriptor/manager.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "internal/remote_descriptor/remote_descriptor.hpp"
+#include "internal/remote_descriptor/storage.hpp"
+
+#include "srf/protos/codable.pb.h"
+
+namespace srf::internal::remote_descriptor {
+
+void Manager::decrement_tokens(std::size_t object_id, std::size_t token_count)
+{
+    DVLOG(10) << "decrementing " << token_count << " tokens from object_id: " << object_id;
```
drobison00, line 29: Is this thread safe, or is it assumed to be running on serial thread?

drobison00 (commented):
Everything looks reasonable, should be no problem to approve after #115 goes in; but, I'd like a description of what these are intended to do, basic overview of an associated workflow, etc..

src/internal/remote_descriptor/manager.cpp
ryanolson: Fixed with a mutex to protect the map.

drobison00 (approved):

PR #141, opened 2022-07-13
Adjust contrib instructions for pip install location

pdmack:
pip install from the build/python dir, not $SRF_HOME/python

Not sure if we need to even mention SRF_PYTHON_INPLACE_BUILD var

@mdemoret-nv


mdemoret-nv (approved):
LGTM

mdemoret-nv:
@gpucibot merge

PR #153, opened 2022-07-27
Add ability to specify port data type for known c++ types from Python

drobison00:
TODO: still need to add additional unit tests, and update to support a 3rd tuple parameter indicating if the raw data type or shared pointer should be used as the underlying node data type.

This will make a number of non-breaking changes to the process of Ingress and Egress port creation via Python.

- Implements @mdemoret-nv 's changes to remove the need for specifying Ingress and Egress port data types during construction, and instead shifts to a system of providing lambda functions that create the appropriate derived object and return a common un-templated base; this results in a much more flexible, cleaner, interface and eliminates any hard cap on the number of ports that can be created on a segment from Python.

- Extends previous updates to support python bindings for ingress/egress port creation to support either simple strings, or for specifying a string + pybind11 registered type tuple.
`pipeline.make_segment("name", ["a", "b", "c"], [], init)`
`pipleine.make_segment("name", [("a", my_type), ("b", another_type), "c"]`

Internally, the "simple string" case, will result in Ingress/Egress nodes with `PyHolder` date types; which can be inefficient, and require multiple edge converters when connected to pure c++ (non-python) data types. With the newly supported syntax, providing a pybind11 registered type will result in an attempt to lookup the associated (wrapped) cpptype and create an Ingress or Egress port using that type.

- The ability to "lookup" constructors for ingress and egress ports via cpptype is accomplished via a new static 'port_registry' class, and a `PortUtil` class. The port registry stores a map from std::type_index to PortUtil, and PortUtil provides a number of helper functions for creating, casting, and manipulating Ingress and Egress ports.


src/public/node/port_registry.cpp
```diff
@@ -0,0 +1,129 @@
...
+    CHECK(names.size() == type_indices.size());
+
+    std::lock_guard<std::recursive_mutex> lock(s_mutex);
+    for (std::size_t idx = 0; idx < names.size(); idx++)
+    {
+        register_name_type_index_pair(names[idx], type_indices[idx]);
+    }
+}
+bool PortRegistry::has_port_util(std::type_index type_index)
+{
```
mdemoret-nv, line 115: Probably want to lock while checking if its registered

src/public/node/port_registry.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+std::shared_ptr<PortUtil> PortRegistry::find_port_util(std::type_index type_index)
+{
+    auto iter_util = PortRegistry::s_registered_port_utils.find(type_index);
+    if (iter_util == PortRegistry::s_registered_port_utils.end())
+    {
+        throw std::runtime_error("No PortUtil registered for type");
+    }
+
+    return iter_util->second;
```
mdemoret-nv, line 127: Also want to lock here

include/srf/segment/builder.hpp
```diff
@@ -95,6 +110,9 @@ class Builder final
     DELETE_COPYABILITY(Builder);
     DELETE_MOVEABILITY(Builder);
 
+    std::shared_ptr<ObjectProperties> get_ingress_dynamic(std::string name, std::type_index type_index);
+    std::shared_ptr<ObjectProperties> get_egress_dynamic(std::string name, std::type_index type_index);
+
```
mdemoret-nv, line 115: Could these just be an overload of `get_egress`? So you would have `get_egress(name)` and `get_egress(name, type_index)`. Ideally, the templated overload would just call the untemplated one passing `typeid(T)` for `type_index`

include/srf/segment/ports.hpp
```diff
@@ -30,24 +30,25 @@ namespace srf::segment {
...
 
-template <template <class> class PortT, typename BaseT, typename... TypesT>
+// template <template <class> class PortT, typename BaseT, typename... TypesT>
+template <typename BaseT>
 class Ports
 {
   public:
-    static constexpr auto PortCount = sizeof...(TypesT);
-    using port_builder_fn_t         = std::function<std::shared_ptr<BaseT>(const SegmentAddress&, const PortName&)>;
+    // static constexpr auto PortCount = sizeof...(TypesT);
```
mdemoret-nv, line 38: Commented code that should be removed

include/srf/segment/ports.hpp
```diff
@@ -30,24 +30,25 @@ namespace srf::segment {
 
 class Definition;
 
-template <template <class> class PortT, typename BaseT, typename... TypesT>
+// template <template <class> class PortT, typename BaseT, typename... TypesT>
```
mdemoret-nv, line 33: Commented code that should be removed

python/srf/_pysrf/include/pysrf/node.hpp
```diff
@@ -173,7 +174,8 @@ namespace pysrf {
 
 template <typename InputT, typename ContextT = srf::runnable::Context>
 class PythonSink : public node::RxSink<InputT, ContextT>,
-                   public pysrf::AutoRegSinkAdapter<InputT>
+                   public pysrf::AutoRegSinkAdapter<InputT>,
+                   public pysrf::AutoRegEgressPort<InputT>
```
mdemoret-nv, line 178: Is there any reason why we wouldnt want to combine `AutoRegSinkAdapter` and `AutoRegEgressPort` into a single object? I cant think of a reason why you would want one but not the other

python/srf/_pysrf/include/pysrf/segment.hpp
```diff
@@ -104,6 +106,40 @@ auto wrap_segment_init_callback(
...
+
+        return (self->*method)(std::forward<const std::string&>(name),
+                               std::forward<pybind11::list>(ingress_port_ids),
+                               std::forward<pybind11::list>(egress_port_ids),
+                               std::forward<decltype(f_wrapped)>(f_wrapped));
+    };
+
+    return func;
+}
+
```
mdemoret-nv, line 142: Do we still need the `std::vector` overload if this exists?

python/srf/_pysrf/include/pysrf/segment.hpp
```diff
@@ -104,6 +106,40 @@ auto wrap_segment_init_callback(
     return func;
 }
 
+/**
+ * @brief
+ * @tparam ClassT
+ * @tparam ArgsT
+ * @param method
+ * @return
+ */
```
mdemoret-nv, line 115: Lets not have autogenerated doxygen comments with no content.

python/srf/core/common.cpp
```diff
@@ -33,6 +61,12 @@ PYBIND11_MODULE(common, m)
            :toctree: _generate
     )pbdoc";
 
+    node::EdgeAdapterRegistry::register_source_adapter(typeid(PyHolder),
+                                                       EdgeAdapterUtil::build_source_adapter<PyHolder>());
+
+    node::EdgeAdapterRegistry::register_sink_adapter(typeid(PyHolder), EdgeAdapterUtil::build_sink_adapter<PyHolder>());
```
mdemoret-nv, line 67: Are these needed. Wont they get registered when a PyHolder node is created?

python/srf/_pysrf/src/segment.cpp
```diff
@@ -181,22 +200,37 @@ std::shared_ptr<srf::segment::ObjectProperties> SegmentProxy::make_sink(srf::seg
...
 }
 
-std::shared_ptr<srf::segment::ObjectProperties> SegmentProxy::get_egress(
-    srf::segment::Builder& self,
-    const std::string& name)
+std::shared_ptr<srf::segment::ObjectProperties> SegmentProxy::get_egress(srf::segment::Builder& self,
+                                                                         const std::string& name)
 {
-   return self.get_egress<PyHolder>(name);
+    auto it_caster = node::PortRegistry::s_port_to_type_index.find(name);
```
mdemoret-nv, line 219: Why is this doing a lookup in a static list of port names to types? There may be a time when we have multiple ports with the same name but different types (like a derived/base class situation).

At this point, the segment proxy should be able to lookup the ingress/egress port object by name and query the type id from there.

include/srf/node/port_registry.hpp
```diff
@@ -0,0 +1,126 @@
...
+    using egress_object_caster_fn_t =
+        std::function<std::shared_ptr<segment::ObjectProperties>(std::shared_ptr<segment::EgressPortBase>)>;
+
+    using ingress_builder_fn_t =
+        std::function<std::shared_ptr<segment::IngressPortBase>(const SegmentAddress&, const PortName&)>;
+    using egress_builder_fn_t =
+        std::function<std::shared_ptr<segment::EgressPortBase>(const SegmentAddress&, const PortName&)>;
+
+    using ingress_tuple_t = std::tuple<ingress_builder_fn_t, ingress_builder_fn_t>;
+    using egress_tuple_t  = std::tuple<egress_builder_fn_t, egress_builder_fn_t>;
```
mdemoret-nv, line 55: Can we add a comment here on why these are all doubled up tuples? It was not clear to me.

python/srf/_pysrf/src/pipeline.cpp
```diff
@@ -18,42 +18,147 @@
...
 
-    static port_type_t build(const std::vector<std::string>& port_ids)
-    {
-        return port_type_t(port_ids);
-    }
-};
-}  // namespace
+using egress_info_t = std::tuple<std::vector<std::string>,
+                                 std::vector<std::type_index>,
+                                 std::vector<srf::node::PortUtil::egress_builder_fn_t>>;
```
mdemoret-nv, line 58: Can we just make these their own struct so its easier to know what each field is.

mdemoret-nv (changes requested):
This is a great new feature and looking really good. Have a few suggestions and I think we can cleanup some pieces by using type information thats already there.

python/srf/_pysrf/src/segment.cpp
drobison00: > Why is this doing a lookup in a static list of port names to types? There may be a time when we have multiple ports with the same name but different types (like a derived/base class situation).

Ryan said this is not allowed by SRF (as in we do not allow ports of the same name and different data types) and not something we need to worry about.

> At this point, the segment proxy should be able to lookup the ingress/egress port object by name and query the type id from there.

I don't believe this is correct given the way things work. The current implementation of `Builder::get_ingress` first calls `get_ingress_base`; which returns an `IngressPortBase`; and is able to up cast it because it has the type information. Absent this information we can't directly cast an IngressPortBase to an ObjectProperties object that we can call `sink_source()` on, because ObjectProperties is a virtual base of the object. To resolve this, I went with storing a name->type_index map.

include/srf/segment/builder.hpp
drobison00: Updated; I don't want to go so far as making the template version call the non-templated version, because they return different types and I don't know all the relevant use cases there; did switch it to an overload though.

python/srf/_pysrf/include/pysrf/node.hpp
drobison00: Probably not; this gave us the most flexibility, but I suppose its not really an issue; I'll combine the inherited object.

python/srf/_pysrf/include/pysrf/segment.hpp
drobison00: No.

python/srf/core/common.cpp
mdemoret-nv: Sounds good.

python/srf/_pysrf/src/segment.cpp
mdemoret-nv: We have discussed this and it will require a larger change to make creating manifolds more explicit

mdemoret-nv (approved):
Once CI is passing this looks good to merge

drobison00:
@gpucibot merge

PR #158, opened 2022-08-10
Load Balancer Update

### LINKED ISSUE #157: [FEA] Relax pe_count for CSP vGPU with hard limits
Azure subscriptions impose a hard limit on vGPU resources in their VM.
So if we run a Morpheus pipeline with 8 threads on a NC6s v2 (6 vCPU) instance we get an abort in thread_engines.cpp

```
F20220809 15:38:51.809185   262 thread_engines.cpp:38] 
Check failed: launch_options().pe_count == m_cpu_set.weight() 
(8 vs. 6) mismatch in the number of cores in the cpu set with respect to the requested pe_count
*** Check failure stack trace: ***
Aborted (core dumped)
```

**Describe the solution(s) you'd like**

- Graceful degradation to the value of m_cpu_set.weight() with appropriate warnings in logging. 
- Possibly make hard v. soft limits configurable via the SRF API and throw the error back up to Morpheus (app layer).

**Describe alternatives you've considered**
Workaround is for the Morpheus pipeline to be run with threads <= vGPU count in Azure VM.

**Additional context**
https://docs.microsoft.com/en-us/azure/databricks/kb/clusters/azure-core-limit

ryanolson:
This PR updates the Load Balancer Manifold from a push based model requiring a node with multiple writers to a pull based one requiring zero runtime resources.

The behavior of the load-balancer is now improved, especially for work-loads for which the individual units of works are asymmetric with respect to compute time. If you had two segments, each assigned to a different class GPU say a T4 and an H100, then the new LB allows for the H100 to consume more work packets as it should complete work faster. The old behavior would blinding assign half the work to one segment and half to the other.

Relates to #157

### DIFF
--- a/ci/iwyu/mappings.imp
+++ b/ci/iwyu/mappings.imp
@@ -97,6 +97,8 @@
 { "symbol": ["nlohmann::json", "private", "<nlohmann/json_fwd.hpp>", "public"] },
 
 # pybind11
+{ "include": [ "<pybind11/detail/common.h>", private, "<pybind11/pybind11.h>", "public" ] },
+
 { "symbol": ["pybind11", "private", "<pybind11/cast.h>", "public"] },
 { "symbol": ["pybind11", "private", "<pybind11/embed.h>", "public"] },
 { "symbol": ["pybind11", "private", "<pybind11/pybind11.h>", "public"] },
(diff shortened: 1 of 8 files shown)


dagardner-nv (approved):
LGTM

mdemoret-nv:
Closing as this PR is very old. Can always reopen if needed.

PR #165, opened 2022-08-22
Update PortBuilder to Work with Types That Do Not Have a Default Constructor

mdemoret-nv:
If you create a source or a sink class that derives from `PythonSink`, `PythonSource` or `PythonNode` and the message type does not have a default constructor, you will run into a compilation error. This was due to the fact that the new Port auto registration added in PR #153 tries to compile ingress/egress ports for both `T` and `shared_ptr<T>`.

This PR adds a check to skip compiling a port with `T` if `T` does not have a default constructor.

### DIFF
--- a/python/srf/_pysrf/include/pysrf/port_builders.hpp
+++ b/python/srf/_pysrf/include/pysrf/port_builders.hpp
@@ -1,6 +1,6 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION &
- * AFFILIATES. All rights reserved. SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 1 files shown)


python/srf/_pysrf/include/pysrf/port_builders.hpp
```diff
@@ -1,6 +1,6 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION &
+ * AFFILIATES. All rights reserved. SPDX-License-Identifier: Apache-2.0
```
ryanolson, line 3: did something happen to your formatter?

drobison00 (approved):

python/srf/_pysrf/include/pysrf/port_builders.hpp
mdemoret-nv: Not sure. I'll run the formatters just to be sure. But CI didnt have any issues with it.

mdemoret-nv:
@gpucibot merge

PR #177, opened 2022-09-26
update abseil, grpc, and ucx versions for cuml compatibility

cwharris:
installing cuml 22.08 along side srf 22.09 does not work, as the abseil, grpc, and ucx versions mismatch. This PR resolves those issues. In doing so, a bug presented itself in how conda build was determining hashes. SRF was attempting to look for a version of libsrf that was not being generated, to fix this, I added abseil as a dependency of SRF, which is a workaround mentioned in https://github.com/conda-forge/arrow-cpp-feedstock/issues/814.


mdemoret-nv (approved):
Heavily tested this PR and added cuml to the test dependencies to guarantee cuml is installable. Approved

mdemoret-nv:
@gpucibot merge

h-vetinari:
Just a note that starting from grpc 1.46 & abseil 20220623, it's not necessary to zip those anymore, or build for several different abseil versions. :)

PR #182, opened 2022-10-07
Switch to github actions

### LINKED ISSUE #154: [FEA] Migrate CI to github actions
Current CI uses Jenkins, which is being phased out in favor of Github actions

Ideally we would use this opportunity to improve CI performance.

dagardner-nv:
In general this should have parity with the existing Jenkins pipeline.

* On pull requests all steps are run except the conda package step
* On pushes to branches named `branch-*` the check script is skipped, and the conda package step is run

It isn't clear if the nightly builds were actually enabled/working. 

Fixes #154


ci/scripts/github/common.sh
```diff
@@ -51,12 +58,18 @@ export DISPLAY_ARTIFACT_URL="${DISPLAY_URL}${ARTIFACT_ENDPOINT}"
...
 #export SCCACHE_LOG=debug
 
-gpuci_logger "Environ:"
-env | sort
+mkdir -p ${WORKSPACE_TMP}
+
+function print_env_vars() {
+    gpuci_logger "Environ:"
+    env | grep -v -E "AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY|TOKEN" | sort
+}
```
mdemoret-nv, line 70: Are these not hidden by the Github UI since they are secrets? This makes me nervious

ci/scripts/github/common.sh
```diff
@@ -37,9 +43,10 @@ export CMAKE_BUILD_WITH_CODECOV="-DCMAKE_BUILD_TYPE=Debug -DSRF_ENABLE_CODECOV=O
 export GIT_DEPTH=1000
 
 # For PRs, $GIT_BRANCH is like: pull-request/989
-REPO_NAME=$(basename "${GIT_URL}" .git)
-ORG_NAME=$(basename "$(dirname "${GIT_URL}")")
-PR_NUM="${GIT_BRANCH##*/}"
+REPO_NAME=$(basename "${GITHUB_REPOSITORY}")
+ORG_NAME="${GITHUB_REPOSITORY_OWNER}"
+PR_NUM="${GITHUB_REF_NAME##*/}"
```
mdemoret-nv, line 48: Will this fail if we arent in a PR? i.e. on a branch build?

ci/scripts/github/conda.sh
```diff
@@ -16,23 +16,11 @@
...
-CONDA_ARGS+=("--label" "${CONDA_PKG_LABEL}")
-CONDA_ARGS="${CONDA_ARGS[@]}" ${SRF_ROOT}/ci/conda/recipes/run_conda_build.sh
-
-gpuci_logger "Archiving Conda Package"
-cd $(dirname ${CONDA_BLD_OUTPUT})
-tar cfj ${WORKSPACE_TMP}/conda_pkg.tar.bz $(basename ${CONDA_BLD_OUTPUT})
-
-gpuci_logger "Pushing results to ${DISPLAY_ARTIFACT_URL}/"
-aws s3 cp ${WORKSPACE_TMP}/conda_pkg.tar.bz "${ARTIFACT_URL}/conda_pkg.tar.bz"
+${SRF_ROOT}/ci/conda/recipes/run_conda_build.sh upload
```
mdemoret-nv, line 26: Wow that simplified things here

mdemoret-nv (commented):
Couple of questions

ci/scripts/github/common.sh
dagardner-nv: Github actions does filter/remove secrets from being printed to the logs, however the docs also say we should avoid printing them in general:
https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets
> Warning: GitHub automatically redacts secrets printed to the log, but you should avoid printing secrets to the log intentionally.

Which is why I added that in, as an extra just-in-case.

ci/scripts/github/common.sh
dagardner-nv: I tested this, on a branch build `PR_NUM` will just be the branch name.
```
(base) root@roci:/SRF# export GITHUB_REF_NAME=branch-22.11
(base) root@roci:/SRF# PR_NUM="${GITHUB_REF_NAME##*/}"
(base) root@roci:/SRF# echo $?
0
(base) root@roci:/SRF# echo $PR_NUM
branch-22.11
```

The `PR_NUM` var is used in two places:

1. As part of the S3 artifact url:
`export ARTIFACT_ENDPOINT="/pull-request/${PR_NUM}/${GIT_COMMIT}/${NVARCH}/${BUILD_CC}"`

1. As part of the check script where we use the github rest api to fetch the changes. 


The check script is skipped on branch builds and I think it works ok as part of the s3 url.

ci/scripts/github/conda.sh
dagardner-nv: Yes :) very much so

ci/scripts/github/common.sh
mdemoret-nv: Ok, we should then add CONDA_TOKEN to the list of envs not to print

ci/scripts/github/common.sh
mdemoret-nv: Sounds good

ci/scripts/github/common.sh
dagardner-nv: Its a regex so it will exclude any env var with "TOKEN" in the name, so this will match on `GH_TOKEN` and `CONDA_TOKEN`

mdemoret-nv: Ah I see

mdemoret-nv (approved):
Looks good now

mdemoret-nv:
@gpucibot merge

PR #183, opened 2022-10-07
Add documentation on how to build the doxygen docs

dagardner-nv:
Looks like VSCode decided to remove two white-space chars.

fixes #127

PR #187, opened 2022-10-10
Fix CI deps

dagardner-nv:
Remove dependency on check so that downstream stages are executed even when the check stage is skipped


CONTRIBUTING.md
```diff
@@ -114,6 +114,13 @@ pip install -e $SRF_HOME/build/python
 pytest $SRF_HOME/python
 ```
 
+### Building API Documentation
+From the root of the SRF repo, configure CMake with `SRF_BUILD_DOCS=ON` then build the `srf_docs` target. Once built the documentation will be located in the `build/docs/html` directory.
+```bash
+cmake -B build -DSRF_BUILD_DOCS=ON .
+cmake --build build --target srf_docs
+```
+
```
mdemoret-nv, line 123: This seems unrelated. Is this intentional?

mdemoret-nv (approved):
One question

CONTRIBUTING.md
dagardner-nv: That is weird! That is the commit from #183 not sure how it got in this pr.

mdemoret-nv: Looks like it just got branched from the wrong commit

mdemoret-nv:
@gpucibot merge

PR #195, opened 2022-10-14
Allow building build without GPU and without a driver

dagardner-nv:
Allows building SRF without a gpu or a driver present when the `SRF_IGNORE_NO_GPU` environment variable is defined.

Currently we build with an image containing the Nvidia driver which pulls in several deps. This change would allow us to build with slimmer images.

### DIFF
--- a/src/internal/system/device_info.cpp
+++ b/src/internal/system/device_info.cpp
@@ -25,6 +25,7 @@
 
 #include <array>
 #include <cstddef>
+#include <cstdlib>
 #include <memory>
 #include <ostream>
 #include <set>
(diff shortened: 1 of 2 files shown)


src/internal/system/device_info.cpp
```diff
@@ -37,20 +37,12 @@ struct NvmlState
...
+        auto nvml_status = nvmlInit_v2();
+        if (nvml_status != NVML_SUCCESS)
         {
-        case NVML_SUCCESS:
-            break;
-        case NVML_ERROR_DRIVER_NOT_LOADED:
-            LOG(WARNING)
-                << "NVML: No NVIDIA GPU driver was detected; setting DeviceCount to 0, CUDA will not be initialized";
+            LOG(WARNING) << "NVML: Error initializing due to " << nvmlErrorString(nvml_status)
+                         << "setting DeviceCount to 0, CUDA will not be initialized";
```
mdemoret-nv, line 44:
```suggestion
            LOG(WARNING) << "NVML: Error initializing due to '" << nvmlErrorString(nvml_status)
                         << "'. Setting DeviceCount to 0, CUDA will not be initialized";
```

mdemoret-nv (approved):
Looks good now.

PR #201, opened 2022-10-20
change pull_request to pull_request_target

jarmak-nv:
The current action uses `on: pull_request` but for it to work successfully, it needs to be `on: pull_request_target`

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target


mdemoret-nv (approved):
Great, was going to suggest this change myself.

mdemoret-nv:
@gpucibot merge

PR #208, opened 2022-11-04
Fix throwing errors from `Executor.join_async()`

mdemoret-nv:
Exceptions from both python and C++ were getting printed to the log but not raised in python when using `Executor.join_async()`. This boils down to the fact that `future::wait()` was called instead of `future::get()` which actually rethrows the exception.

Added some tests to check this functionality

### DIFF
--- a/python/srf/tests/utils.cpp
+++ b/python/srf/tests/utils.cpp
@@ -1,0 +1,51 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


python/srf/tests/utils.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <pybind11/cast.h>
+#include <pybind11/pybind11.h>
+
+#include <stdexcept>
+
+namespace srf::pytests {
+
+namespace py = pybind11;
+
+PYBIND11_MODULE(utils, m)
```
drobison00, line 29: Can we use 'module' instead of m, the short variable names generate style warnings.

drobison00 (approved):

python/srf/tests/utils.cpp
mdemoret-nv: Done.

dagardner-nv (approved):
LGTM

mdemoret-nv:
@gpucibot merge

PR #209, opened 2022-11-04
Draft: Testing out codecov config options

mdemoret-nv:
Testing PR


mdemoret-nv:
rerun tests

tests/test_channel.cpp
```diff
@@ -41,230 +41,230 @@
...
-    const CopyMoveCounter& const_r_value = CopyMoveCounter();
-    channel->await_write(const_r_value);
-
-    channel->await_read(std::ref(output));
-    // Old value was copied
-    check_counter(const_r_value, 0, 0, true, false);
-    // New value move in and move out
-    check_counter(output, 1, 2, false, false);
-}
+// class TestChannel : public ::testing::Test
```
drobison00, line 44: Is this intentional?

mdemoret-nv: No, was just trying to force a lot of coverage changes

mdemoret-nv:
Current findings before putting this on hold until 23.01:

- The difference between the coverage locally vs the coverage reported in codecov.io is significant
   - Locally, I see the following:
      - Running all of the C++ tests: 52.0% of lines (3768/7244) and 19.4% of branches (3872/19935)
      - Running the python tests: 51.5% of lines (3715/7216) and 18.5% of branches (3694/19935)
      - Running the both C++ and Python (combined): 63.9% of lines (4657/7289) and 24.2% of branches (4817/19935)
- The `codecov` binary actually runs `gcov` itself and uploads that report instead of the specified filename
   - To disable this you need the `-X gcov` command line option
- When running `gcovr` you need to specify the `gcov` executable that matches the compiler
   - Since we compile using binaries in conda, our compiler is `x86_64-conda_cos6-linux-gnu-g++` so we need to specify the gcov binary as `x86_64-conda-linux-gnu-gcov`.
   - You can do this using the option `--gcov-executable=x86_64-conda_cos6-linux-gnu-g++`
- When running `gcovr` you want the timestamps of all of the .o files to match the corresponding .gcno files
   - This means we likely need to run a clean build and the tests for coverage all in the same stage
   - It might actually simplify the stages a bit since we will not need 2 artifacts for both debug and release builds
   - This was the main reason this PR was shelved since it would require large changes to the CI system
- Its important to specify flags in the codecov.yml file
   - Look at this PR for examples
   - Adding flags enables more features from Github (its a bit confusing why that is)

mdemoret-nv:
Closing in favor of #233

PR #218, opened 2022-11-21
Distributed Runtime

### LINKED ISSUE #120: [FEA] Add convenient methods to EncodedObject
EncodedObject should add new methods for common patterns.

This should either update the behavior of `add_memory_block` or provide a new method that does the following:

- takes a buffer_view `(addr, size, kind)
- checks to see if the `addr` is registered with the data plane
- if so, make a remote memory descriptor from the view,
- if not, allocate a buffer, copy the data to the buffer and make a remote md to the newly allocated buffer.
- optionally, we could decided to registered the unregistered buffer if the size is sufficiently large to avoid a copy.

It should be noted, that this method should only be called for creating a contiguous block of memory in the encoded object of a sufficiently large size to warrant a remote memory descriptor.

For smaller objects, we might find we should use a ReusablePool of buffers and used packed descriptors into those buffers.  The reusable pool offers an optimization over just allocating a buffer from the registered memory resource.

ryanolson:
This PR is a major step toward the completion Multi-Node or fully distributed SRF pipelines.

This PR puts in place the framework for the following components:
- control plane
- data plane
- fiber based grpc
- ucx runtime

This PR improves upon the following existing components:
- codable protocol

This PR add the initial public runtime and resources interface.

Closes #120


include/srf/pubsub/subscriber.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * [T] -> EncodedObject<T> -> EncodedStorage -> RemoteDescriptor -> Transient Buffer -> Data Plane Tagged Send
+ *
+ * Subscriber<T> Data Path:
+ * Data Plane Tagged Received -> Transient Buffer -> RemoteDescriptor -> Subscriber/Source<T> ->
+ *
+ * Subscriber<RemoteDescriptor> Data Path:
+ * Data Plane Tagged Received -> Transient Buffer -> RemoteDescriptor -> Subscriber/Source<RemoteDescriptor> ->
+ */
+template <typename T>
+class Subscriber final : public node::Queue<T>,
```
mdemoret-nv, line 53: If this inherits from `public node::Queue<T>` that means you can make connections that write to the subscriber. i.e. `node::make_edge(source, subscriber)`. This needs to inherit from something else I think.

include/srf/runtime/api.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include "srf/codable/api.hpp"
+#include "srf/pubsub/forward.hpp"
+
+#include <cstdint>
+#include <memory>
+
+namespace srf::runtime {
+
+class IPartition;
+class IRemoteDescriptorManager;
```
mdemoret-nv, line 29: These should be in "forward.hpp"

include/srf/runtime/api.hpp
```diff
@@ -0,0 +1,129 @@
...
+     */
+    virtual std::shared_ptr<pubsub::ISubscriberService> make_subscriber_service(const std::string& name) = 0;
+
+    // Friends
+
+    template <typename T>
+    friend class pubsub::Publisher;
+
+    template <typename T>
+    friend class pubsub::Subscriber;
```
mdemoret-nv, line 126: This relationship seems weird to me. I feel like the public runtime objects shouldnt have friends. Let alone ones in a completely different namespace. All of the methods on these objects should be part of the public API.

include/srf/types.hpp
```diff
@@ -41,18 +41,15 @@ using Mutex = userspace_threads::mutex;  // NOLINT(readability-identifier-naming
 
 using CondV = userspace_threads::cv;  // NOLINT(readability-identifier-naming)
 
-using InstanceID = std::uint32_t;                      // NOLINT(readability-identifier-naming)
-using MachineID  = std::uint32_t;                      // NOLINT(readability-identifier-naming)
-using Instances  = std::map<InstanceID, std::string>;  // NOLINT(readability-identifier-naming)
+using InstanceID = std::uint64_t;                      // NOLINT(readability-identifier-naming)
```
mdemoret-nv, line 44: I added TagID here which helped distinguish it from InstanceID.

protos/CMakeLists.txt
```diff
@@ -23,44 +23,62 @@ add_library(srf_architect_protos)
...
-    PROPERTIES
-      CXX_CLANG_TIDY ""
-      C_INCLUDE_WHAT_YOU_USE ""
-      CXX_INCLUDE_WHAT_YOU_USE ""
-      EXPORT_COMPILE_COMMANDS OFF
+  PROPERTIES
+  CXX_CLANG_TIDY ""
+  C_INCLUDE_WHAT_YOU_USE ""
+  CXX_INCLUDE_WHAT_YOU_USE ""
+  EXPORT_COMPILE_COMMANDS OFF
```
mdemoret-nv, line 63: All of the indentation here got messed up.

python/srf/_pysrf/include/pysrf/plugins.hpp
```diff
@@ -17,6 +17,29 @@
...
+{
+    void on_data(ProgressEvent&& event) final
+    {
+        auto* promise = static_cast<boost::fibers::promise<bool>*>(event.tag);
+        promise->set_value(event.ok);
+    }
+};
+
+}  // namespace srf::internal::rpc
+========
```
mdemoret-nv, line 42: Merge error

src/internal/pubsub/publisher.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <unordered_map>
+#include <utility>
+
+namespace srf::internal::runtime {
+class Partition;
+}
+
+namespace srf::internal::pubsub {
+
+class Publisher : public Base,
```
mdemoret-nv, line 42: Can we call this `PublisherService` to distinguish it from the public `Publisher`?

src/internal/pubsub/subscriber.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include <string>
+#include <utility>
+
+namespace srf::internal::pubsub {
+
+/**
+ * @brief The internal type-erased Subscriber
+ *
+ */
+class Subscriber final : public Base,
```
mdemoret-nv, line 43: Also call this `SubscriberService`

include/srf/pubsub/publisher.hpp
```diff
@@ -0,0 +1,148 @@
...
+    // [Ingress<T>] publish T by capturing it as an encoded object, then pushing that encoded object to the internal
+    // publisher
+    channel::Status await_write(T&& data) final;
+
+    void await_start() final
+    {
+        // form a persistent connection to the operator
+        // data flowing in from operator edges are forwarded to the public await_write
+        m_persistent_channel = std::make_unique<srf::node::SourceChannelWriteable<T>>();
+        srf::node::make_edge(*m_persistent_channel, *this);
```
mdemoret-nv, line 80: I dont love this. Why is the `m_persistent_channel` needed here?

mdemoret-nv (approved):
Approving with comments.

include/srf/codable/api.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+class IStorage;
+class IEncodableStorage;
+class IDecodableStorage;
+
+template <typename T>
+class Encoder;
+
+template <typename T>
+class Decoder;
```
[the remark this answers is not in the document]
ryanolson: I think we could do that as long as we properly annotate as private with IWYU. 

I've been using forward.hpp as the thing you use from another directory, and using explicit forward declares for the files in the directory. 

If you include forward.hpp in the definition header, I think IWYU makes you include that everywhere even though you don't need it.

include/srf/concepts/invocable.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <concepts>
+#include <functional>
+
+namespace srf::concepts {
+
+template <typename F, typename... ArgsT>
+concept invocable = requires(F&& f, ArgsT&&... args)
+{
+    std::invoke(std::forward<F>(f), std::forward<ArgsT>(args)...);
+};
```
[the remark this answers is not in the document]
ryanolson: Sneaky. Yeah, it's not used it got included accidentally. That should be in the cpp20 branch

include/srf/memory/buffer_view.hpp
```diff
@@ -41,6 +41,22 @@ class const_buffer_view  // NOLINT
     const_buffer_view()          = default;
     virtual ~const_buffer_view() = default;
 
+    const_buffer_view(const const_buffer_view& other) = default;
+    const_buffer_view& operator=(const const_buffer_view&) = default;
+
+    const_buffer_view(const_buffer_view&& other) noexcept :
+      m_data(other.m_data),
+      m_bytes(other.m_bytes),
+      m_kind(other.m_kind)
```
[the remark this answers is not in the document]
ryanolson: Stylistically I agree, but fundamental types don't have move constructors. Moves degrade to copies in that case.

include/srf/memory/codable/buffer.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "srf/memory/buffer_view.hpp"
+#include "srf/memory/memory_kind.hpp"
+
+#include <type_traits>
+#include <typeindex>
+
+namespace srf::codable {
+
+template <typename T>
+struct codable_protocol<T, std::enable_if_t<std::is_same_v<T, srf::memory::buffer>>>
```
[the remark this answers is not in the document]
ryanolson: Yeah, that was just a carry over from using a is base of...

include/srf/node/operators/operator.hpp
```diff
@@ -74,7 +74,9 @@ std::shared_ptr<channel::Ingress<T>> Operator<T>::channel_ingress()
     {
         return ingress;
     }
+    LOG(INFO) << "before shared_from_this()";
     auto this_operator = this->shared_from_this();
+    LOG(INFO) << "after shared_from_this()";
```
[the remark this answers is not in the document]
ryanolson: Good spot. That was where I was getting an exception throw when inheriting multiple level of enable shared from this.

include/srf/node/operators/unique_operator.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#pragma once
+
+#include "srf/channel/ingress.hpp"
+#include "srf/node/sink_properties.hpp"
+
+namespace srf::node {
+
+template <typename T>
+class UniqueOperator : public SinkProperties<T>
```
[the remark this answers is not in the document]
ryanolson: Nor I. It's just a place holder until we get the new operator semantics. 

It's an operator that doesn't inherit from enable shared from this.

include/srf/pubsub/subscriber.hpp
ryanolson: The new operator semantic should give us the option to make this a channel provider and not appear to have a dangling upstream connection. 

In fact, the upstrsm of the queue is connected to the internal pipeline that provides either a T or a RemoteDescriptkr.

include/srf/pubsub/publisher.hpp
ryanolson: Publisher is both directly writable and upstream connectable. 

Because of that, we can not just immediate request a control plane drop service on this tag. 

We have to wait to make the control plane drop request until after all upstream have been disconnected AND a call to request stop have been made on the object. 

We use are using the persistent channel model to do this. Initialize the persistent channel.  Upstream ingress acceptors can connect. Request stop drops the persistent connection. When the last connection is released the control plane messsge is issued.

protos/CMakeLists.txt
ryanolson: Where is that formatting triggered? Am I missing an extension?

include/srf/runtime/api.hpp
ryanolson: It's primary to limit the things that are public on the IPartition object. 

make_xxx_service are not really meant to be publicly callable, but need to be called by the create method on the publisher/subscriber. 

If there is a better way, im open to suggestions

src/internal/pubsub/publisher.hpp
ryanolson: Done

src/internal/pubsub/subscriber.hpp
ryanolson: Done

include/srf/concepts/invocable.hpp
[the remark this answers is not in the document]
ryanolson: Removed

python/srf/_pysrf/include/pysrf/plugins.hpp
ryanolson: Fixed.

include/srf/types.hpp
ryanolson: Done

include/srf/node/operators/operator.hpp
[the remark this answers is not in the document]
ryanolson: fixed

include/srf/node/operators/broadcast.hpp
```diff
@@ -17,61 +17,67 @@
...
+    {
+        auto typed_ingress = std::dynamic_pointer_cast<channel::Ingress<T>>(ingress);
+
+        CHECK(typed_ingress) << "Invalid ingress type passed to broadcast";
+
+        m_output_channels.push_back(typed_ingress);
     }
 
-    std::vector<SourceChannelWriteable<T>> m_downstream_channels;
+    // std::vector<SourceChannelWriteable<T>> m_downstream_channels;
```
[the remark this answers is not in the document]
ryanolson: fixed

include/srf/memory/codable/buffer.hpp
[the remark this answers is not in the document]
ryanolson: Fixed

cmake/dependencies.cmake
```diff
@@ -116,9 +118,9 @@ find_path(CUB_INCLUDE_DIRS "cub/cub.cuh"
 # =========
 rapids_find_package(gRPC REQUIRED
   GLOBAL_TARGETS
-    gRPC::address_sorting gRPC::gpr gRPC::grpc gRPC::grpc_unsecure gRPC::grpc++ gRPC::grpc++_alts gRPC::grpc++_error_details gRPC::grpc++_reflection
-    gRPC::grpc++_unsecure gRPC::grpc_plugin_support gRPC::grpcpp_channelz gRPC::upb gRPC::grpc_cpp_plugin gRPC::grpc_csharp_plugin gRPC::grpc_node_plugin
-    gRPC::grpc_objective_c_plugin gRPC::grpc_php_plugin gRPC::grpc_python_plugin gRPC::grpc_ruby_plugin
+  gRPC::address_sorting gRPC::gpr gRPC::grpc gRPC::grpc_unsecure gRPC::grpc++ gRPC::grpc++_alts gRPC::grpc++_error_details gRPC::grpc++_reflection
+  gRPC::grpc++_unsecure gRPC::grpc_plugin_support gRPC::grpcpp_channelz gRPC::upb gRPC::grpc_cpp_plugin gRPC::grpc_csharp_plugin gRPC::grpc_node_plugin
+  gRPC::grpc_objective_c_plugin gRPC::grpc_php_plugin gRPC::grpc_python_plugin gRPC::grpc_ruby_plugin
```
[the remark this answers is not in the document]
mdemoret-nv: Ok, lets figure this out in the rename/refactor

include/srf/codable/api.hpp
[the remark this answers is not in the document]
mdemoret-nv: >If you include forward.hpp in the definition header, I think IWYU makes you include that everywhere even though you don't need it.

Interesting. Ok with these for now then.

include/srf/memory/buffer_view.hpp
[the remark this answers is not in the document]
mdemoret-nv: Fair enough. Then this needs to function more like the assignment operator and use `std::swap()` so that `other` is left in an uninitialized state.

protos/CMakeLists.txt
mdemoret-nv: We will sort out the format settings in another PR

include/srf/pubsub/publisher.hpp
mdemoret-nv: After discussion, I'm fine with this for now.

include/srf/node/operators/unique_operator.hpp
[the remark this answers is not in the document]
mdemoret-nv: We can address after the non-linear pipeline stuff is merged.

include/srf/pubsub/subscriber.hpp
mdemoret-nv: We can address after the non-linear pipeline stuff is merged.

include/srf/runtime/api.hpp
mdemoret-nv: This is fine for now.

include/srf/runtime/api.hpp
mdemoret-nv: We can address after IWYU changes are merged.

mdemoret-nv:
@gpucibot merge

PR #224, opened 2022-11-28
  Adds missing functionality to allow dynamic modules to forward their configuration to nested children.

### LINKED ISSUE #223: [FEA]: Add the ability to get the current module's config file when initializing a python defined module.
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Critical (currently preventing usage)

### Please provide a clear description of problem this feature solves

Currently, we have to capture a python function in order to do python module initialization, this function is stored on the PySegmentModule derived class, but does not have direct access to the derived class.

In the same way that this `py_segment_init` function does not have direct access to the `register_module_input` and `register_module_output` calls on the the PySegmentModule, neither does it have direct access to the config. This is a problem, because in the case of nested python modules we may want to forward pieces of the current module's configuration to modules that it creates.

Since the builder keeps track of the initializing module stack, it can directly reference the currently executing module to retrieve its configuration, in the same way that it proxies calls to register module input/output.

### Describe your ideal solution

Two new API calls need to be added; one to the Builder, and one to the pysrf BuilderProxy

```
const nlohmann::json Builder::get_current_module_config() const { ... }
```

```
const pybind11::dict BuilderProxy::get_current_module_config(srf::segment::Builder& self) const {
   auto json_config = self.get_module_config();
   
   return cast_from_json(json_config);
}
```

### Describe any alternatives you have considered

_No response_

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow SRF's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/nv-morpheus/SRF/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

drobison00:
Adds missing functionality to allow dynamic modules to forward their configuration to nested children.

Closes #223

### DIFF
--- a/include/srf/segment/builder.hpp
+++ b/include/srf/segment/builder.hpp
@@ -199,6 +199,12 @@ class Builder final
     void register_module_input(std::string input_name, std::shared_ptr<segment::ObjectProperties> object);
 
     /**
+     * Get the json configuration for the current module under configuration.
+     * @return nlohmann::json object.
+     */
+    const nlohmann::json get_current_module_config();
+
+    /**
      * Register an output port on the given module -- note: this in generally only necessary for dynamically
      * created modules that use an alternate initializer function independent of the derived class.
      * See: PythonSegmentModule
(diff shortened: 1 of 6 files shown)


mdemoret-nv (approved):

drobison00:
@gpucibot merge

PR #237, opened 2022-12-01
C++ Reorganization

ryanolson:
- [x] create cpp dir
- [x] move include, src, test and benchmarks to cpp/
- [x] update cmake


mdemoret-nv (commented):
Did we want to prepare for the `mre` library by putting everything in a `cpp/mrc` folder? So it would look like:

- `cpp`
   - `CMakeLists.txt`
   - `mrc`
      - `CMakeLists.txt`
      - `benchmark`
      - `include`
      - `src`
      - `test`

This way when we add `mre`, then it would be:

- `cpp`
   - `CMakeLists.txt`
   - `mre`
      - `CMakeLists.txt`
   - `mrc`
      - `CMakeLists.txt`

It would add one extra directory layer now but could save us some refactoring in the future. On the other hand, we will need to do a lot of refactoring if we make the MRE library anyways. I could go either way on this.

ryanolson:
> Did we want to prepare for the `mre` library by putting everything in a `cpp/mrc` folder? So it would look like:
> 
> * `cpp`
>   
>   * `CMakeLists.txt`
>   * `mrc`
>     
>     * `CMakeLists.txt`
>     * `benchmark`
>     * `include`
>     * `src`
>     * `test`
> 
> This way when we add `mre`, then it would be:
> 
> * `cpp`
>   
>   * `CMakeLists.txt`
>   * `mre`
>     
>     * `CMakeLists.txt`
>   * `mrc`
>     
>     * `CMakeLists.txt`
> 
> It would add one extra directory layer now but could save us some refactoring in the future. On the other hand, we will need to do a lot of refactoring if we make the MRE library anyways. I could go either way on this.

done

mdemoret-nv (approved):

ryanolson:
@gpucibot merge

PR #239, opened 2022-12-02
C++20 version of upcoming C++23 std::expected

ryanolson:
- removes `tl::expected`
- adds c++20 implemetnation from https://github.com/RishabhRD/expected


cpp/mrc/include/mrc/core/std23_expected.hpp
```diff
@@ -0,0 +1,1476 @@
+/**
```
mdemoret-nv, line 1: Does this need a header for the source location

cpp/mrc/src/internal/expected.hpp
```diff
@@ -32,7 +31,8 @@ enum class ErrorCode
 
 class Error;
 
-using UnexpectedError = tl::unexpected<Error>;  // NOLINT
+// todo(#219) - update tidy to allow the following typedef
```
mdemoret-nv, line 34: Is this still needed?

mdemoret-nv (approved):
Just a couple of questions

cpp/mrc/include/mrc/core/std23_expected.hpp
ryanolson: done

PR #241, opened 2022-12-02
MRC changes related to utility consolidation

drobison00:
Depends on: https://github.com/nv-morpheus/utilities/pull/1/files#diff-9a3c07061febeba4d3e5191e87706e262644abf8514317eeb084e67137e7c807

Shifts to using a shared utilities submodule for shared cmake, scripts, utilities, etc.. This shared repo provides a number of new morpheus_utils_xxx cmake macros/functions that are now used throughout the configuration process.


cwharris:
Should this be a draft pr?

CMakeLists.txt
```diff
@@ -27,41 +27,45 @@ option(MRC_BUILD_DOCS "Enable building of API documentation" OFF)
...
-# Default to using "NATIVE" for CUDA_ARCHITECTURES to build based on GPU in system
-if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
-  set(CMAKE_CUDA_ARCHITECTURES "NATIVE")
-  message(STATUS "CMAKE_CUDA_ARCHITECTURES was not defined. Defaulting to '${CMAKE_CUDA_ARCHITECTURES}' to build only for local architecture. Specify -DCMAKE_CUDA_ARCHITECTURES='ALL' to build for all archs.")
-endif()
+# Configure project package manager
+morpheus_utils_initialize_package_manager(
+    MRC_USE_CONDA
+    BUILD_SHARED_LIBS
+    MRC_VCPKG_TOOLCHAIN
```
ryanolson, line 57: Is VCPKG still a viable and tested option?

If not, should we remove it?

CMakeLists.txt
```diff
@@ -27,41 +27,45 @@ option(MRC_BUILD_DOCS "Enable building of API documentation" OFF)
...
+morpheus_utils_initialize_cuda_arch(mrc)
 
 project(mrc
-  VERSION 22.11.00
-  LANGUAGES C CXX
-)
-
+    VERSION 22.11.00
+    LANGUAGES C CXX
+    )
```
ryanolson, line 68: We need to sync on CMake formatters.

We recently added a config for the VS Code extension CMake Language Support to allow free-form indenting which was adding a lot of diffs.

Is this a CLion formatter? If so, we should get our VS Code extensions and your CLion config playing nice together.

cmake/load_morpheus_utils.cmake
```diff
@@ -0,0 +1,49 @@
...
+function(find_and_configure_morpheus_utils version)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "morpheus_utils")
+
+  set(fetchcontent_tmp "${FETCHCONTENT_BASE_DIR}")
+  set(FETCHCONTENT_BASE_DIR "${CMAKE_SOURCE_DIR}/.cache") # default location
+  if (${CPM_SOURCE_CACHE})
+    set(FETCHCONTENT_BASE_DIR "${CPM_SOURCE_CACHE}")
+  endif()
+
+  FetchContent_Declare(
```
ryanolson, line 32: Here is where we need to have a discussion on FetchContent vs Git Submodules.

I don't regularly touch files in the cmake directory, but with a submodule that is initialized and updated, it feels like a folder which I can navigate around and possibly edit.

With FetchContent, a copy is fetched and put in the build path. I think you can still edit that local copy and it maintains its git origins so you can push PRs back to the source should any changes be made.

My preference here would be a submodule, but I think it's something we need to evaluate the pros and cons and make a group decision.

CMakeLists.txt
drobison00: That's a good question. It would be really nice to be able to rip all the VCPKG code out.

cmake/load_morpheus_utils.cmake
drobison00: I don't think I have any strong opinion here, although I'm more or less on board with using sub-modules since it reduces dependencies.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "morpheus_utils"]
+	path = external/utilities
+	url = git@github.com:nv-morpheus/utilities.git
+	branch = devin-issue-225-87
```
drobison00, line 4:
```suggestion
	url = https://github.com/nv-morpheus/utilities.git
	branch = branch-23.01
```

drobison00: Roll before merge.

cmake/dependencies.cmake
```diff
@@ -138,12 +119,11 @@ rapids_find_package(nlohmann_json REQUIRED
...
 # prometheus
 # =========
-set(PROMETHEUS_CPP_VERSION "1.0.0" CACHE STRING "Version of Prometheus-cpp to use")
-include(deps/Configure_prometheus)
+morpheus_utils_configure_prometheus_cpp(${PROMETHEUS_CPP_VERSION})
 
 # libcudacxx
 # =========
-include(deps/Configure_libcudacxx)
+morpheus_utils_configure_libcudacxx(${LIBCUDACXX_VERSION})
```
mdemoret-nv, line 126: Why are these two different and pass variables into the functions? We should have them all be the same.

mdemoret-nv (approved):
Just a couple of suggestions.

CMakeLists.txt
```diff
@@ -91,12 +92,16 @@ if(NOT DEFINED CMAKE_CUDA_HOST_COMPILER)
...
+
   endif()
 endif()
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
   # Our version of NVCC officially only supports clang versions 3.2 - 13, we are now using 14
   set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -allow-unsupported-compiler")
+
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcoroutines-ts")
+  #add_compile_options(-fcoroutines-ts) # Required for clang to use the coroutines libraries
```
mdemoret-nv, line 104: Which one of these are we using? `CMAKE_CXX_FLAGS` or `add_compile_options`?

ci/scripts/github/checks.sh
```diff
@@ -23,6 +23,8 @@ fetch_base_branch
 update_conda_env
 
 rapids-logger "Configuring CMake"
+git submodule init
+git submodule update
```
mdemoret-nv, line 27: Why do we use:
```
git submodule update --init --recursive
```
in some places and
```
git submodule init
git submodule update
```
in others? I think I would prefer the single line option: `git submodule update --init --recursive`

python/mrc/_pymrc/src/tracers.cpp
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TODO(Devin)
+#include "pymrc/tracers.hpp"
```
mdemoret-nv, line 19: Outstanding TODO

python/mrc/benchmarking/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 list(APPEND CMAKE_MESSAGE_CONTEXT "benchmarking")
 
-mrc_add_pybind11_module(watchers SOURCE_FILES watchers.cpp)
-mrc_add_pybind11_module(tracers SOURCE_FILES tracers.cpp trace_statistics.cpp)
+mrc_add_pybind11_module(watchers SOURCE_FILES watchers.cpp trace_statistics.cpp)
+#mrc_add_pybind11_module(tracers SOURCE_FILES tracers.cpp trace_statistics.cpp)
```
mdemoret-nv, line 18: Commented code

python/mrc/benchmarking/__init__.py
```diff
@@ -13,10 +13,10 @@
...
-from .tracers import sync_tracing_state
-from .tracers import trace_channels
-from .tracers import trace_operators
 from .watchers import LatencyWatcher
 from .watchers import ThroughputWatcher
+from .watchers import get_tracing_stats
+from .watchers import reset_tracing_stats
+from .watchers import sync_tracing_state
+from .watchers import trace_channels
+from .watchers import trace_operators
```
mdemoret-nv, line 22: Are these changes supposed to be here?

python/mrc/benchmarking/watchers.cpp
```diff
@@ -28,16 +29,46 @@
...
+    });
+    LatencyTracer.def("emit", &latency_tracer_t::emit);
+
+    /**
+     * @brief ThroughputTracer
+     */
+    auto ThroughputTracer =
+        py::class_<throughput_tracer_t, std::shared_ptr<throughput_tracer_t>>(m, "ThroughputTracer");
+    ThroughputTracer.def(py::init<std::size_t>());
+
```
mdemoret-nv, line 71: This seems like it should be in a separate PR

mdemoret-nv (approved):
Approved but it looks like there are a few things that shouldnt be in there regarding tracers

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "morpheus_utils"]
 	path = external/utilities
-	url = git@github.com:nv-morpheus/utilities.git
-	branch = devin-issue-225-87
+	url = https://github.com/nv-morpheus/utilities.git
```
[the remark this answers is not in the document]
drobison00: Cool; I didn't even know you could do this.

CMakeLists.txt
drobison00: The uncommented version; will update.

ci/scripts/github/checks.sh
drobison00: Good catch, I prefer the single line as well, will fix.

python/mrc/benchmarking/__init__.py
drobison00: Yeah, I moved a few things out of the Tracers module that were causing problems with pybind stubs, so they're in watchers.

python/mrc/benchmarking/watchers.cpp
drobison00: It can't be, it breaks pybind11 stubs.

PR #244, opened 2022-12-05
C++20 Coroutines

ryanolson:
Implementation based on the work of jbaldwin/libcoro.

The reason for the fork is due to the fact that we will need to touch every awaitable object to apply a MRC context. The other major difference is ThreadPool/Scheduler/Executor discovery and a new SchedulingPolicy that will enable us to replicate boost::fiber behaviors in some critical places where coroutines will eventually replace fibers.

Blocked by #239


mdemoret-nv (approved):
Dont see anything wrong with this. Havent gone over everything since coroutines is still new to me.

ryanolson:
@gpucibot merge

PR #246, opened 2022-12-05
Remove draconian regex filter that eliminates entire tests from being run

### LINKED ISSUE #238: [FEA]: Enable test_mrc_private.x
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

High

### Please provide a clear description of problem this feature solves

The internal tests found in `test_mrc_private.x` are disabled.

```
# Issues:
# * test_mrc_private - https://github.com/nv-morpheus/MRC/issues/33
ctest --output-on-failure \
      --exclude-regex "test_mrc_private|nvrpc" \
      --output-junit ${REPORTS_DIR}/report_ctest.xml
```

`ctest` does not have the ability to excluded specific tests like the `--gtest_filter` option.

### Describe your ideal solution

Re-enable these important tests

We can continue to use `ctest` and the current regex exclude, but we should also add a second explicit launch of the `test_mrc_private.x` executable with any individual test excluded.

### Describe any alternatives you have considered

_No response_

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow MRC's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/nv-morpheus/MRC/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

ryanolson:
Resolves #238 

TODOs:
- [x] add a new macro to detect if running in CI
- [x] enable the skipping of some tests that don't work in the CI env


cpp/mrc/tests/test_mrc.hpp
```diff
@@ -26,13 +26,23 @@
...
 namespace mrc {
+
+inline void skip_if_in_ci()
+{
+    if (std::getenv("CI") != nullptr)
+    {
+        GTEST_SKIP() << "Test skipped in CI";
+    }
+}
+
```
mdemoret-nv, line 45: Why is this a function and `SKIP_IF_IN_CI` one a macro?

ci/scripts/github/test_codecov.sh
```diff
@@ -44,10 +44,9 @@ set +e
 # Tests known to be failing
 # Issues:
 # * test_mrc_private - https://github.com/nv-morpheus/MRC/issues/33
-# * nvrpc - https://github.com/nv-morpheus/MRC/issues/34
-ctest --output-on-failure \
-      --exclude-regex "test_mrc_private|nvrpc" \
-      --output-junit ${REPORTS_DIR}/report_ctest.xml
+CODE_COV_RUN=y ctest --output-on-failure \
```
mdemoret-nv, line 47: This will give inconsistent results when running code coverage locally vs in CI.

cpp/mrc/src/tests/common.hpp
```diff
@@ -17,10 +17,30 @@
...
+    }
+
+#define SKIP_IF_CODE_COV()                          \
+    {                                               \
+        if (std::getenv("CODE_COV_RUN") != nullptr) \
+        {                                           \
+            GTEST_SKIP() << "Test skipped in CI";   \
+        }                                           \
+    }
+
```
mdemoret-nv, line 43: Since this variable is only set in the CI scripts, this will give different coverage results when running locally vs in CI. Instead, since we enable code coverage with the `MRC_ENABLE_CODECOV` CMake option, we should use a `#define` directive set by CMake. You would need an option set in the libmrc CMakeLists.txt:

```cmake
if (MRC_ENABLE_CODECOV)
  target_compile_definitions(libmrc INTERFACE "MRC_CODECOV_ENABLED")
endif()
```

And then the macro becomes:
```cpp
#ifdef MRC_CODECOV_ENABLE
    #define SKIP_IF_CODE_COV() GTEST_SKIP() << "Skipping test when code coverage is enabled";
#else
    #define SKIP_IF_CODE_COV()
#endif
```

mdemoret-nv (changes requested):
Few suggestions. You also are need some changes in `cmake/setup_coverage.cmake`

cpp/mrc/src/tests/common.hpp
ryanolson: good call

cpp/mrc/tests/test_mrc.hpp
ryanolson: i forgot i had put it in test_mrc.hpp first, then realized we didn't need it in the public test.

ryanolson: i added the updated macro to the public tests as well in case we need it.

mdemoret-nv (approved):

mdemoret-nv:
@gpucibot merge

Issue #247, opened 2022-12-05
[BUG]: failing control plane tests


ryanolson:
### Version

23.01

### Which installation method(s) does this occur on?

Docker

### Describe the bug.

Bug appears to be a control plane client/server shutdown problelm.

The tests appear to be failing on the resource construction.

### Minimum reproducible example

```shell
./test_mrc_private.x -v 10 --gtest_filter=TestCodable.Double --gtest_repeat=-1
```

### Relevant log output

```shell
[ RUN      ] TestCodable.Double
I20221205 19:48:48.760887  1829 topology.cpp:221] topology restricted to cpu_set: 0-63
I20221205 19:48:48.760946  1829 partitions.cpp:67] forming memory and device partitions
I20221205 19:48:48.760970  1829 partitions.cpp:147] detected 1 nvidia gpus; 1 host cpu/memory domains
I20221205 19:48:48.760994  1829 partitions.cpp:179] host cpu/memory domain: 0 has 64 logical cpus (0-63) with 503.8 GiB memory
I20221205 19:48:48.761031  1829 partitions.cpp:188] host resources will be split into 1 partitions across 32 cores
I20221205 19:48:48.761073  1829 partitions.cpp:239] host_partition_id: 0 contains 64 logical cpus (0-63) with 503.8 GiB memory
I20221205 19:48:48.761104  1829 partitions.cpp:257] assigning cuda_device_id: 0; pcie: 00000000:02:00.0 to host_partition_id: 0
I20221205 19:48:48.761126  1829 partitions.cpp:269] evaluating engine factory cpu sets for host_partition 0-63
I20221205 19:48:48.761171  1829 fiber_manager.cpp:41] creating fiber task queues on 64 threads
I20221205 19:48:48.761193  1829 fiber_manager.cpp:42] thread_binding :  TRUE
I20221205 19:48:48.761211  1829 fiber_manager.cpp:43] memory_binding :  TRUE
I20221205 19:48:48.783917  1829 manager.cpp:67] building runnable/launch_control resources on host_partition: 0
I20221205 19:48:48.784070  1829 manager.cpp:84] building ucx resources for partition 0
I20221205 19:48:48.784101  1829 resources.cpp:43] constructing network resources for partition: 0 on partitions main task queue
I20221205 19:48:48.798871  1829 launch_control.hpp:192] preparing engines using engine factory default; pe_count=1; engines_per_pe: 1
[...]

PR #248, opened 2022-12-05
Fix potential race condition

ryanolson:
Prior to this PR, there was a possible race condition that could be hit by two execution paths where,

- path 1 calls Server::drop_stream which is a yielding call
- path 2 call Server:m_connections->drop_all_streams();

`Server::drop_stream` locks the server's state, but `Server::m_connections` is not protected by that mutex.

Relates to #247

### DIFF
--- a/cpp/mrc/src/internal/control_plane/server.cpp
+++ b/cpp/mrc/src/internal/control_plane/server.cpp
@@ -588,6 +588,7 @@ void Server::drop_stream(writer_t& writer)
 {
     const auto stream_id = writer->get_id();
     drop_stream(stream_id);
+    writer.reset();
 }
 
 void Server::drop_stream(const stream_id_t& stream_id)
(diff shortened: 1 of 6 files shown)


cpp/mrc/src/internal/control_plane/server.cpp
```diff
@@ -169,7 +169,7 @@ void Server::do_service_await_join()
 {
     // clear all instances which drops their held stream writers
     DVLOG(10) << "awaiting all streams";
-    m_connections.drop_all_streams();
```
ryanolson, line 172: If this PR solves this part of the problem, we should remove `drop_all_streams` from the connection manangers interface.

ryanolson:
Note, this will pass CI, because CI is still skipping test_mrc_private.x.

ryanolson:
We could also convert the Server's

```
mutable boost::fibers::mutex m_mutex;
```

to a 

```
mutable boost::fibers::recursive_mutex m_mutex;

mdemoret-nv (approved):

ryanolson:
@gpucibot merge

PR #253, opened 2022-12-09
Make devcontainer use lightweight Dockerfile and mount conda env

cwharris:
Replaces the devcontainer's full ci-based image with a lightweight development image which builds faster. Devcontainer adjusted to mount the conda environment, such that changes to the environment yaml will not require a rebuild of the development image.

Also deletes post-start script, as git username and email address is inherited from the host environement.


.devcontainer/devcontainer.json
```diff
@@ -1,41 +1,28 @@
...
-    // },
+
     "runArgs": [
         "--gpus=all",
-        "--runtime=nvidia",
-        "--network=host",
-        "--cap-add=SYS_PTRACE",
-        "--cap-add=SYS_NICE",
-        "--security-opt",
-        "seccomp=unconfined"
```
ryanolson, line 26: sys_nice is needed for huge pages
sys_ptrace and seccomp=unconfied is needed for debug
network=host is needed for some ucx transports

ryanolson:
I think we need to make a distinction between a "core developer's" container and a development container.

Currently, I don't think devcontainers supports multiple definition with different Dockerfiles, but it is an issue that I have subscribed to for updates. Once that is supported, we can have multiple variants of devcontainers.

For those of us actively contributing to MRC, I think the current container image makes the most sense, since it allows core contributors to debug and replicate the state of the CI runs.

I think for developers looking for a playground to kick the tires on MRC, that is where a more streamlined container like this might come into play. 

I certainly would like to see a devcontainer that is generally compatible with codespaces and that can be used virtually anywhere.

cwharris:
@drobison00 , you're a

cwharris:
@drobison00 as the other core contributor, what do you think?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ghcr.io/trxcllnt/devcontainer-images/cmake-ninja-sccache-llvm-cuda-mambaforge:latest AS base
```
trxcllnt, line 1:
```suggestion
FROM ghcr.io/trxcllnt/devcontainer-images/cmake-ninja-sccache-llvm-cuda-mambaforge:latest
```

.devcontainer/devcontainer.json
```diff
@@ -1,41 +1,28 @@
...
+        "dockerfile": "Dockerfile"
     },
-    // todo(252)
-    // new capability with vs code 1.74 (November 2022 Release)
-    // https://github.com/microsoft/vscode-docs/blob/main/remote-release-notes/v1_74.md#gpu-host-requirement
-    // enable this after more testing is conducted
-    // "hostRequirements": {
-    //     "gpu": true
-    // },
+
```
trxcllnt, line 6: Doing this will let you get rid of `--gpus=all` and `--runtime=nvidia` in the `runArgs`:

```suggestion
"hostRequirements": { "gpu": true },
```

.devcontainer/devcontainer.json
```diff
@@ -45,15 +32,19 @@
...
+                "ms-vscode.cpptools",
                 "ms-vscode.test-adapter-converter",
                 "twxs.cmake",
                 "xaver.clang-format"
             ]
         }
+    },
+
+    "settings": {
+        "C_Cpp.intelliSenseEngine": "disabled",
```
trxcllnt, line 47: This will disable the MS C++ extension, but it doesn't look like the clangd extension is installed to replace it?

Issue #176, opened 2022-09-22
[BUG] Extension libraries compiled with pybind11 don't have cpptype visibility via get_internals


drobison00:
**Describe the bug**

If we register a new pybind11._class object in an external module, it will store its `type_info` / `tinfo` struct in its own copy of internal registered python types.

If an object created in this way is passed to another independently compiled library (in this case libsrf_pysrf), subsequent calls to `get_internals().registered_types_py[ object type ptr ]` won't find the expected type `type_info` structure.

Since we can't rely on this behavior, we need to attach the wrapped cpptype as an out of band attribute for the class object. This is necessary so that we can match the internal PortAdapter type_id with the type_id of the wrapped object.

PR #261, opened 2022-12-20
Non-Linear pipelines and Non-Runnable Nodes

mdemoret-nv:
## The Goal

The goal of this PR was to enable non-linear pipelines from python code and to also allow the user to create nodes that do not have a progress engine. For example:

```python

# Create a C++ source
source = m.SourceDerived(seg, "source")

# Make a broadcast node
broadcast = mrc.core.node.Broadcast(seg, "broadcast")

# Create a C++ sink with a progress engine
sink = m.SinkBase(seg, "sink")

def on_next(x):
   print("Got: {}".format(type(x)))

# Create a Python sink without a progress engine
sink_component = seg.make_sink_component("sink_component", on_next, None, None)

# Connect source -> broadcast
seg.make_edge(source, broadcast)

# Connect broadcast -> sink
seg.make_edge(broadcast, sink)

# Connect broadcast -> sink_component
seg.make_edge(broadcast, sink_component)
```

The example above shows creating a pipeline with a single source and multiple readers and is interesting for a few reasons:
 - The `sink_component` object will have its `on_next` function run immediately, using the progress engine from the `source` node
 - The `sink` object will run it's body asynchronously on a different thread from any other node
 - The C++(Source) -> Broadcast -> C++(Sink) path should not need the python GIL

While it was possible to do something similar to this in C++ code, this was not possible in Python and presented a few challenges.

## The Problem

Currently, there are several node types in C++ that do not have a progress engine (`Queue`, `Muxer`, `Broadcast`, etc). These object were referred to as "Operators" in the code but for the remainder of this PR, they will be referred to as "Components".

### Problem 1 - C++ Broadcast Node Types

Simply wrapping the existing Component nodes in python bindings doesn't really solve our goal. If you wanted to connect two C++ nodes together through a broadcast from Python, you would need to derive from `Broadcast<T>` for your particular C++ type, create python bindings for this single broadcast object, and then create that object in python. You would need to repeat this for every C++ type you plan to use in your pipeline. Since the C++ broadcast does very little with the C++ object, this is incredibly tedious for the amount of work it accomplishes.

### Problem 2 - Implementing Source/Node/Sink Components

The above example used a simple Sink Component, but what about a Source Component? A Sink Component is easy to reason about since it's a push-driven design. The upstream progress engine calls `await_write` when it is done with a value, and we can easily have this call the Sink component `on_next` function. But a Source Component would need to be a pull driven system similar to the existing `Queue`. How would this work? Can you connect a pull-driven node to a push-driven one? Our current node/edge system didn't have any way to distinguish between the two.

### Problem 3 - Coupling of Channels in the Node Inheritance Tree

The number of base node types has grown significantly and it can be difficult to know which one to derive from when making a new node type. For example, this is the full inheritance for an `RxSink`:

 - `RxSink<T>`
   - `RxSinkBase<T>`
     - `SinkChannel<T>`
       - `SinkChannelBase<T>`
       - `SinkProperties<T>`
         - `SinkPropertiesBase`
       - `ChannelAcceptor<T>`
     - `Watchable`
   - `RxRunnable<ContextT>`
     - `runnable::RunnableWithContext<ContextT>`
       - `Runnable`
     - `RxSubscribable`
   - `RxPrologueTap<T>`

It can be difficult to know which ones to inherit from because they are very similar in functionality and naming. For example, what's the difference between a `SinkChannel<T>`, `SinkChannelBase<T>`, `SinkProperties<T>`, and `ChannelAcceptor<T>`?

Additionally, many of these classes are tighly coupled with `Channel<T>` objects such as `SinkChannel<T>`, `SinkChannelBase<T>`, and `ChannelAcceptor<T>`. Early iterations of this PR that attempted to use the existing classes were very awkward because Components do not often use channels. Even the previous `Operator` class duplicated bits of functionality from `SinkChannelBase` to get around this issue.

## The Solution

To solve these problems, the clever `ChannelProvider`/`ChannelAcceptor` system was expanded on. In this system, nodes can declare themselves as a `ChannelProvider` if they give a reference to their channel to another node and a `ChannelAcceptor` if they can accept a reference to the channel of another node. But instead of accepting and providing channel object, we accept and provide ingress/egress objects.

 - `Ingress` - An object that you can write to, i.e. it has an `await_write(T)` method
 - `Egress` - An object that you can read from, i.e. it has an `await_read(T&)` method
 - `Provider` - A node provides objects to other nodes to hold onto
 - `Acceptor` - A node accepts and holds onto other node objects

This leaves us with the 4 possible node object types:

 - `IngressProvider` - I give other nodes an Ingress object that they can write to
 - `IngressAcceptor` - I accept an Ingress object from other nodes that I can write to
 - `EgressProvider` - I give other nodes an Egress object that they can read from
 - `EgressAcceptor` - I accept an Egress object from other nodes that I can read from

With these 4 classes, we can describe any node as a combination of 1 or more of the above objects. For example, if a node is a push-based source with a progress engine, it should derive from `IngressAcceptor`. We can apply this logic to all properties of nodes to come up with the handy table below which maps node concepts to these 4 classes.

|  | `IngressProvider` | `IngressAcceptor` | `EgressProvider` | `EgressAcceptor` |
| --- | :-: | :-: | :-: | :-: |
| Is a Source?  |  | `X` | `X` |  |
| Is a Sink?  | `X` |  |  | `X` |
| Push-Based? | `X` | `X` |  |  |
| Pull-Based? |  |  | `X` | `X` |
| Requires Progress Engine? |  | `X` |  | `X` |

Finally, if all nodes derive from these objects, then it's pretty simple to determine valid connections with the following overloads:

- `make_edge(IngressAcceptor<T> source, IngressProvider<T> sink)`
  - Push-based edge
- `make_edge(EgressProvider<T> source, EgressAcceptor<T> sink)`
  - Pull-based edge

How does this help with the problems above? These classes help in the following ways:
1. It makes it much easier to distinguish between nodes with progress engines from those that do not. i.e. a Sink vs a Sink Component
2. It establishes a simple system for determining if edges should be push or pull based and eliminates invalid connections between incompatible nodes
3. It moves most of the edge connection logic from the nodes, to the Ingress/Egress objects that are shared between nodes (Allowing us to do things like deferring edge conversion for C++ nodes)
4. It separates the node inheritance tree from Channels

### An Example

It's easier to understand this concept with some examples. Lets look at some of the properties of the `RxSink`. Anything we can answer "No" to eliminates options from the above chart:

 - Does it have a progress engine? Yes
   - If it did not have a progress engine, we could eliminate both `Provider` classes
 - Is it a Sink? Yes
   - If it wasn't a sink, we could have eliminated `IngressProvider` and `EgressAcceptor`
 - Is it a Source? No
   - This eliminates `IngressAcceptor` and `EgressProvider`

This leaves only two options:

 - `class RxSink: public IngressProvider, public EgressAcceptor`
   - `IngressProvider` - The `RxSink` provides an Ingress object to upstream nodes. 
     - When upstream nodes call `await_write` on the Ingress object, the value is written into the `RxSink`'s channel.
     - When the `RxSink`'s progress engine is ready for a new value, it calls `await_read` on the channel object to get the next value for processing
   - `EgressAcceptor` - The `RxSink` can also run in pull-mode and receives an Egress object from upstream nodes.
     - In this situation, there is no channel on the `RxSink`
     - When the `RxSink`'s progress engine is ready for a new value, it calls `await_read` on the Egress object. This method will then "pull" the next value from the upstream node for processing.

Let's consider a harder example, the `Queue` object which converts a push based upstream node, to a pull based downstream node.

 - Does it have a progress engine? No
   - The `Queue` only wraps a channel object and does not have a progress engine
   - Eliminates `EgressAcceptor` and `IngressAcceptor`
 - Is it a Sink? Yes
   - Upstream nodes can write directly into the channel
 - Is it a Source? Yes
   - Downstream nodes can read from the channel

This leaves the following options:

 - `class Queue: public IngressProvider, public EgressProvider`
   - `IngressProvider` - The `Queue` provides an Ingress object to upstream nodes.
     - When upstream nodes call `await_write` on the Ingress object, the value is written into the `Queues`'s channel.
   - `EgressProvider` - The `Queue` provides an Egress object to downstream nodes.
     - When downstream nodes are ready for their next value, `await_read` is called which pulls the next value from the `Queue`'s channel.

Below is a listing of all node types and what type of node they are:

|  | `IngressProvider` | `IngressAcceptor` | `EgressProvider` | `EgressAcceptor` |
| --- | :-: | :-: | :-: | :-: |
| `RxSource`  |  | `X` | `X` |  |
| `GenericSourceComponent`*  |  |  | `X` |  |
| `RxNode` | `X` | `X` | `X` | `X` |
| `RxNodeComponent` | `X` | `X` |  |  |
| `RxSink` | `X` |  |  | `X` |
| `RxSinkComponent` | `X` |  |  |  |
| `Broadcast` | `X` | `X` |  |  |
| `Queue` | `X` |  | `X` |  |
| `CombineLatest` | `X**` | `X` |  |  |
| `Conditional` | `X` | `X**` |  |  |
| `Muxer` | `X` | `X` |  |  |
| `Router` | `X` | `X**` |  |  |
| `WritableSubject` |  | `X` |  |  |
| `ReadableSubject` |  |  |  | `X` |

`*` - An `RxSourceComponent` isn't possible since Reactive is push based by design.

`**` - These are `Multi-` variants that accept multiple edge connections specified by a key

### Making an Edge Connection

Edges can only be made between the following two pairs:
   - `IngressAcceptor <- IngressProvider`
   - `EgressProvider -> EgressAcceptor`
  
Because nodes can derive from multiple different classes at the same time, it's possible for both overloads to be valid. In this case, the `Ingress` overload will be preferred to continue backwards compatibility. If no overload is valid, then an edge cannot be made. Again, this is best illustrated with some examples:

 - `RxSource -> RxSink`
   - `IngressAcceptor <- IngressProvider` is valid (will be preferred)
   - `EgressProvider -> EgressAcceptor` is valid
 - `Queue -> RxSink`
   - `EgressProvider -> EgressAcceptor` is valid
 - `WritableSubject -> RxSink`
   - `IngressAcceptor <- IngressProvider` is valid
 - `Queue -> RxSinkComponent`
   - No overload is valid. Cannot connect.
 - `RxNodeComponent -> RxSinkComponent`
   - `IngressAcceptor <- IngressProvider` is valid

Which overload is used will be completely hidden from the node implementation and does not alter the node functionality in any way.

## Implementation

In the current system, all edges are formed by the sink providing an `Ingress<T>` to the source. This only supports push-based edge connections. Pull-based edges are simulated by loaning out a channel to another progress engine. So how do we support pull-based edges? What are these Ingress/Egress objects? And how does evertyhing shut down properly with both push and pull based edges?

### The `EdgeHolder` and `EdgeHandle` Classes

The two classes that make everything work are the `EdgeHolder` and the `EdgeHandle`. The `EdgeHolder` class functions very similarly to the existing `SinkChannel` class but it is separated from the concept of a Channel. It's responsible for maintaining the lifetime of an `EdgeHandle` with a shared pointer. When a dependent edge connection is formed, the shared pointer to the `EdgeHandle` is released and a `weak_ptr` is kept. Any additional connections will try to reuse the `EdgeHandle` held by the `weak_ptr`. Once all dependent `shared_ptr` to the `EdgeHandle` are released, the `EdgeHandle` is destroyed and the node is stopped.

This introduces the concept of "dependent" nodes. "Dependent" nodes are nodes who derive from `IngressProvider` or `EgressProvider` and have made an edge connection using that base class. "Dependent" nodes are not allowed to shut down until all of their independent nodes have released their edge connection. This is because the dependent node is not the driver of when it will be called. For example, if you have a SourceComponent connected to an `RxSink`, the SourceComponent is the dependent node and the `RxSink` is the independent. If you think about this situation another way, the `RxSink` is the independent node because it is free to call `await_read` on the SourceComponent's Egress object whenever it wants. So even though data flows from the SourceComponent to the `RxSink`, the SourceComponent must remain alive as long as the `RxSink` can pull values from it. To signal to the downstream node that it should shutdown, the `closed` return value of `await_read` should be used.

This was tricky to get correct because it is important to shut down nodes in the correct order. If any `Acceptor` class is still alive, the corresponding `Provider` class must stay alive as well. To guarantee this, on destruction, the `EdgeHolder` will generate an error if any outstanding Ingress/Egress objects are still pending.

### Connected and Disconnected Events

On the `EdgeHandle` class, two types of events can be registered. On connection and on disconnection. These events help simplify the edge connection process and make it easier to guarantee the lifetime of objects. For example, to call `on_complete` when upstream connections have all been dropped, the `ForwardingIngressProvider` uses this code:

```cpp
auto inner_edge = std::make_shared<ForwardingEdge>(*this);

inner_edge->add_disconnector([this]() {
   // Only call the on_complete if we have been connected
   this->on_complete();
});

IngressProvider<T>::init_owned_edge(inner_edge);
```

The "connected" event fires when an edge connection is made for the first time. It only fires when a complete edge has been formed and only fires the first time. The disconnected event is the opposite and only fires when the last edge has been released. These events are very useful to guarantee specific actions occur at the right time. Such as calling `on_complete`, or releasing downstream edge connections.

## Differences

This section lists the differences between the old system and the new system to make it easy to upgrade.

### Concepts

 - Pushing functionality from classes into Ingress/Egress objects
    - One major difference from before is much more logic for a node is contained in the Ingress/Egress objects themselves, instead of inside of the nodes. 
    - TBD more info on this.

### Class Name Changes

This is a table of the old class names and what should be used in their place:

| Old Class | New Class | Notes |
| --- | --- | :-- |
| `Operator<T>` | `NodeComponent<T>` | `Operator` was too overloaded of a term and functionally identical to a Node Component |
| `SourceChannelWriteable<T>` | `WritableSubject<T>` | "Subjects" are bookends to an edge chain. If you need to manually call `await_write` use the `WritableSubject` |
| `UniqueOperator<T>` | N/A | This object was a temporary fix and has been replaced |
| `SourceProperties` | `IIngressAcceptor`/`IEgressProvider` | What type of source needs to be specified |
| `SinkProperties` | `IIngressProvider`/`IEgressAcceptor` | What type of sink needs to be specified |
| `SourceChannel`/`SourceChannel` | N/A | These classes still exist but should only be derived from and not used as function arguments or with `shared_ptr` |
| `OperatorComponent<T>` | `LambdaNodeComponent<T>` | Functionally identical. Consistent naming |

### Remaining Items
- [ ] Cleanup the file organization
   - [ ] Move the classes in `channel_holder.hpp` into separate files
   - [ ] Move specific node types into their own namespace, i.e. The Rx classes, Components, etc.
- [ ] Rename `IEdgeWritable` and `IEdgeReadable` after Ingress/Egress to better work with the other names
- [ ] Finish improving the gRPC, subscription and control plane pieces to use more of the new nodes


cmake/setup_coverage.cmake
```diff
@@ -66,7 +66,7 @@ if(MRC_ENABLE_CODECOV)
     EXCLUDE ${CODECOV_REPORT_EXCLUSIONS}
     EXECUTABLE "ctest"
     EXECUTABLE_ARGS "--exclude-regex"
-    EXECUTABLE_ARGS "'test_srf_private|nvrpc'"
+    EXECUTABLE_ARGS "'test_mrc_private|nvrpc'"
```
ryanolson, line 69: we no longer exclude either of these.  nvrpc is gone and test_mrc_private should be run.

cpp/mrc/include/mrc/core/utils.hpp
```diff
@@ -31,36 +31,6 @@ namespace mrc {
 
 std::uint16_t hash(const std::string&);
 
-template <typename T>
-constexpr auto type_name() noexcept
-{
-    std::string_view name = "[with T = <UnsupportedType>]";
-    std::string_view prefix;
```
ryanolson, line 38: suffix and prefix should be moved to the _MSC_VER block as no other block needs them.

cpp/mrc/include/mrc/manifold/composite_manifold.hpp
```diff
@@ -65,7 +73,7 @@ class CompositeManifold : public Manifold
     }
 
   private:
-    void do_add_input(const SegmentAddress& address, node::SourcePropertiesBase* input_source) final
+    void do_add_input(const SegmentAddress& address, node::IIngressAcceptorBase* input_source) final
```
ryanolson, line 76: Base or Handle for the fully type-erased class.  In channel we have IngressHandle.

probably time for consistency.

I thin Base for classes that provide function and Handle for those that are used an opaque "handles".

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1059 @@
...
+    std::shared_ptr<T> shared_from_this()
+    {
+        return std::shared_ptr<T>(virtual_enable_shared_from_this_base::shared_from_this(), static_cast<T*>(this));
+    }
+    std::shared_ptr<const T> shared_from_this() const
+    {
+        return std::shared_ptr<const T>(virtual_enable_shared_from_this_base::shared_from_this(),
+                                        static_cast<const T*>(this));
+    }
+};
```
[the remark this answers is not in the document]
mdemoret-nv: This isnt used anymore and I planned on removing it unless you think it has value.

ryanolson:
In reviewing this PR, one thing stands out, which we can work to fix in the future.

When you implement any node-level object with a specific function/operation, you have already made the decision on the ownership of the incoming or outgoing channel, by the nature if that node-level object being defined from Ingress/Egress Provider/Acceptor.

But the specific operation that you are implementing should not care about channel ownership, it should know only that it depends on 0 or more input channels and 0 or more output channels.

I propose in the `next` implementation on coroutines, we further separate the definition of the operation that the node is performing from the specific ownership of the channel.

Furthermore, we can then bake more information into `Edge` at the graph node level, such that we should be able to use the edge connections to traverse the graph.

cpp/mrc/include/mrc/segment/builder.hpp
```diff
@@ -354,6 +379,13 @@ class Builder final
   private:
     using sp_segment_module_t = std::shared_ptr<mrc::modules::SegmentModule>;
 
+    // template <typename SinkNodeT>
```
drobison00, line 382: Delete

cpp/mrc/src/internal/pubsub/publisher_service.cpp
```diff
@@ -48,16 +48,16 @@ PublisherService::PublisherService(std::string service_name, runtime::Partition&
   m_runtime(runtime)
 {}
 
-channel::Status PublisherService::publish(mrc::runtime::RemoteDescriptor&& rd)
-{
-    return this->await_write(std::move(rd));
-}
+// channel::Status PublisherService::publish(mrc::runtime::RemoteDescriptor&& rd)
```
drobison00, line 51: Delete

cpp/mrc/src/public/node/edge_builder.cpp
```diff
@@ -31,68 +34,91 @@
...
+    // Start dynamic lookup
+    VLOG(2) << "Looking for edge adapter: (" << type_name(target_type.full_type()) << ", "
+            << type_name(ingress->get_type().full_type()) << ")";
+    VLOG(2) << "- (" << target_type.full_type().hash_code() << ", " << ingress->get_type().full_type().hash_code()
+            << ")";
+
+    // Loop over the registered adaptors
+    const auto& adaptors = EdgeAdapterRegistry::registered_ingress_adapters;
+
+    for (const auto& adapt : adaptors)
```
drobison00, line 101: Mirror of my comment in EdgeRegistry: can we utilize a double map here instead of iterating through all possible adapters, or is there something that makes this not possible? Maybe even a multi-map for each source type, so we're not testing so many possibilities.

cpp/mrc/src/public/node/edge_registry.cpp
```diff
@@ -29,7 +31,7 @@ namespace mrc::node {
 
 void EdgeRegistry::register_converter(std::type_index writer_type, std::type_index reader_type, build_fn_t converter)
 {
-    VLOG(2) << "Registering converter for " << writer_type.hash_code() << " " << reader_type.hash_code();
+    // VLOG(2) << "Registering converter for " << type_name(writer_type) << " " << type_name(reader_type);
```
drobison00, line 34: Can we just change the VLOG level here, if its too verbose, instead of commenting this out.

cpp/mrc/src/tests/test_grpc.cpp
```diff
@@ -182,7 +183,7 @@ TEST_F(TestRPC, StreamingServerWithHandler)
 
     auto stream  = std::make_shared<stream_server_t>(service_init, m_resources->partition(0).runnable());
     auto handler = std::make_unique<ServerHandler>();
-    handler->enable_persistence();
+    // handler->enable_persistence();
```
drobison00, line 186: Delete?

cpp/mrc/src/tests/test_main.cpp
```diff
@@ -20,19 +20,24 @@
...
 
+#include <csignal>
 #include <stdexcept>
 
-// NOLINTNEXTLINE(readability-identifier-naming)
-__attribute__((noreturn)) void TestFailuresThrowExceptions()
-{
-    throw std::runtime_error("exception rather than std::abort");
-}
+// __attribute__((__noreturn__)) void test_failures_throw_exceptions()
```
drobison00, line 28: Delete?

cpp/mrc/tests/test_main.cpp
```diff
@@ -19,17 +19,19 @@
 
 #include <gflags/gflags.h>  // for ParseCommandLineFlags
 #include <gtest/gtest.h>    // IWYU pragma: keep
+#include <stdnoreturn.h>
 
-// NOLINTNEXTLINE(readability-identifier-naming)
-// void TestFailuresThrowExceptions()
+// __attribute__((__noreturn__)) void test_failures_throw_exceptions()
```
drobison00, line 24: Delete?

python/mrc/_pymrc/include/pymrc/edge_adapter.hpp
```diff
@@ -59,141 +61,94 @@ namespace mrc::pymrc {
...
+            if (!typed_ingress)
+            {
+                // Cant do anything about this ingress
+                return std::shared_ptr<node::IngressHandleObj>(nullptr);
+            }
+
+            auto ingress_type = ingress_handle->get_type();
+
+            // Check to see if we have a conversion in pybind11
+            if (pybind11::detail::get_type_info(ingress_type.unwrapped_type(), false))
```
drobison00, line 94: This actually ends up being problematic, because the internal data structure that Pybind uses is specific to the PYBIND11_INTERNALS_ID value, meaning if you call this in Morpheus for a type that was registered in MRC, it may not find it and you'll always fall back to the Pyobject case (https://github.com/nv-morpheus/MRC/issues/176). We're probably going to want an outside registry lookup to map to the correct pybind11 object.

Check pybind/detail/internal.h:270
```c++
#define PYBIND11_INTERNALS_ID "__pybind11_internals_v" \
    PYBIND11_TOSTRING(PYBIND11_INTERNALS_VERSION) PYBIND11_INTERNALS_KIND PYBIND11_COMPILER_TYPE PYBIND11_STDLIB PYBIND11_BUILD_ABI PYBIND11_BUILD_TYPE "__"

#define PYBIND11_MODULE_LOCAL_ID "__pybind11_module_local_v" \
    PYBIND11_TOSTRING(PYBIND11_INTERNALS_VERSION) PYBIND11_INTERNALS_KIND PYBIND11_COMPILER_TYPE PYBIND11_STDLIB PYBIND11_BUILD_ABI PYBIND11_BUILD_TYPE "__"

/// Each module locally stores a pointer to the `internals` data. The data
/// itself is shared among modules with the same `PYBIND11_INTERNALS_ID`.
inline internals **&get_internals_pp() {
    static internals **internals_pp = nullptr;
    return internals_pp;
}
```

python/mrc/_pymrc/include/pymrc/node.hpp
```diff
@@ -244,6 +290,44 @@ class PythonSource : public node::RxSource<OutputT, ContextT>,
...
+template <typename OutputT>
+class PythonSourceComponent : public node::LambdaSourceComponent<OutputT>,
+                              public pymrc::AutoRegSourceAdapter<OutputT>,
+                              public pymrc::AutoRegIngressPort<OutputT>
+{
+    using base_t = node::LambdaSourceComponent<OutputT>;
+
+    using base_t::base_t;
+
+    //   public:
```
drobison00, line 302: Delete?

python/mrc/_pymrc/include/pymrc/port_builders.hpp
```diff
@@ -157,8 +158,8 @@ struct PortBuilderUtil
 
         if (!mrc::node::PortRegistry::has_port_util(type_idx))
         {
-            VLOG(2) << "Registering PyMRC port util for: " << type_name<port_type_t>() << " "
-                    << "=> " << type_name<port_dtype_t>() << " " << type_idx.hash_code();
+            // VLOG(2) << "Registering PyMRC port util for: " << type_name<port_type_t>() << " "
```
drobison00, line 161: Would prefer this to be uncommented and put at a higher VLOG level if its too verbose; its very helpful in tracking down issues where we get a PyObject edge instead of an optimized edge.

python/mrc/_pymrc/include/pymrc/utilities/function_wrappers.hpp
```diff
@@ -0,0 +1,412 @@
...
+    using cpp_fn_t       = std::function<ReturnT(ArgsT...)>;
+    using return_t       = std::conditional_t<std::is_same<ReturnT, void>::value, pybind11::detail::void_type, ReturnT>;
+    using function_ptr_t = ReturnT (*)(ArgsT...);
+
+    // Default construct with an empty object. Needed by pybind11 casters
+    PyFuncHolder() = default;
+
+    // PyFuncHolder(cpp_fn_t&& cpp_fn, std::string repr) : m_cpp_fn(std::move(cpp_fn)), m_repr(std::move(repr)) {}
+
+    // explicit PyFuncHolder(pybind11::function&& fn)
```
drobison00, line 91: Delete

python/mrc/_pymrc/include/pymrc/utilities/function_wrappers.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+        if (!isinstance<function>(src))
+        {
+            return false;
+        }
+
+        auto func = reinterpret_borrow<function>(src);
+
+        // /*
+        //    When passing a C++ function as an argument to another C++
```
drobison00, line 288: Delete

python/mrc/_pymrc/src/utilities/object_wrappers.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+#include "mrc/exceptions/runtime_error.hpp"
+
+namespace mrc::pymrc {
+
+namespace py = pybind11;
+
+PyObjectWrapper::PyObjectWrapper(pybind11::object&& to_wrap) : m_obj(std::move(to_wrap)) {}
+
+PyObjectWrapper::~PyObjectWrapper()
```
drobison00, line 28: Do we guarantee that this object goes out of scope before the python interpreter exits?

drobison00 (changes requested):
Looks good for the most part; I have a few naming issues that I pointed out which I think would be good to address for clarity and readability.

Also, would be nice to have a comment block on all the new classes giving a brief indicator of their function and expected use.

ryanolson (approved):

python/mrc/_pymrc/include/pymrc/port_builders.hpp
mdemoret-nv: I'll test out higher VLOG levels. This code gets called before the logging can get initialized and generates a warning from glog stating that logging was performed before logging was set up.

python/mrc/_pymrc/include/pymrc/edge_adapter.hpp
mdemoret-nv: I agree with you that we will need to address this issue but this code existed before and the functionality hasnt changed. Only the file has been changed and the location of where this lookup happens.

Seems like we will likely need to do something similar to the Automatic type registration that we did for ports.

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1041 @@
...
+    // Allow edge builder to call set_edge
+    friend EdgeBuilder;
+
+    // Allow multi edge holder to access protected elements
+    template <typename KeyT, typename OtherT>
+    friend class MultiEdgeHolder;
+};
+
+template <typename KeyT, typename T>
+class MultiEdgeHolder
```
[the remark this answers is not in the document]
mdemoret-nv: How would that work? `MultiEdgeHolder` requires a key type. For a single `EdgeHolder` there is no key type. Not sure how this would work. Open to ideas

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1041 @@
...
+              (!m_is_deferred && m_full_type.has_value() && m_unwrapped_type.has_value()))
+            << "Inconsistent deferred setting with concrete types";
+    }
+
+    std::optional<std::type_index> m_full_type;       // Includes any wrappers like shared_ptr
+    std::optional<std::type_index> m_unwrapped_type;  // Excludes any wrappers like shared_ptr if they exist
+    bool m_is_deferred{false};                        // Whether or not this type is deferred or concrete
+};
+
+struct EdgeHandleObj
```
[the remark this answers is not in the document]
mdemoret-nv: It started off as a simple struct with 2 properties but got more complicated. Agreed it should be a class.

`Obj` is short for object and is just a way to differentiate the various class names. We have:

- `EdgeLifetime`
   - Holds a function object that gets called when destroyed. Can be armed or disarmed
- `EdgeTag`
   - Typeless base class for all edges. Holds onto `EdgeLifetime` objects. Maintains some state about the edge (whether or not its connected or disconnected)
- `EdgeHandle`
   - Typed `EdgeTag`
- `EdgeTypePair`
   - Type info about the edge including the wrapped and unwrapped types
- `EdgeHandleObj`
   - Object that holds onto a `EdgeTag` and `EdgeTypePair`.
   - Passed between `EdgeHolder` classes
   - Holds onto an edge to allow edge connection but prevents users from directly interacting with edges (i.e. we dont want users calling `await_write` directly on edge objects. They must be set to an `EdgeTag` 
- `EdgeHolder`
   - Primary class for holding onto edge objects
   - Maintains lifetime of edge objects
   - Base class for all nodes

Jotting these down for now. Open to ideas for different names

cpp/mrc/include/mrc/node/edge_channel.hpp
```diff
@@ -0,0 +1,125 @@
...
+                VLOG(10) << "Closing channel from EdgeChannelReader";
+            }
+
+            m_channel->close_channel();
+        }
+    }
+
+    virtual channel::Status await_read(T& t)
+    {
+        return m_channel->await_read(t);
```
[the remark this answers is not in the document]
mdemoret-nv: No, it should be impossible that `m_channel` is null. Added a check to guarantee this.

cpp/mrc/include/mrc/node/edge_channel.hpp
```diff
@@ -0,0 +1,125 @@
...
+            {
+                VLOG(10) << "Closing channel from EdgeChannelWriter";
+            }
+            m_channel->close_channel();
+        }
+    }
+
+    virtual channel::Status await_write(T&& t)
+    {
+        return m_channel->await_write(std::move(t));
```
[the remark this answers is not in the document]
mdemoret-nv: `EdgeChannel` cannot be made with an empty channel. It must wrap a non-null object.

python/mrc/_pymrc/src/utilities/object_wrappers.cpp
mdemoret-nv: No we dont. This object exists currently in MRC. Would be better to address this in a separate issue

cpp/mrc/include/mrc/node/channel_holder.hpp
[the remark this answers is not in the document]
drobison00: I don't think 'Obj' is helpful, because, all of these types are Objects, so adding 'Obj' to some makes it confusing as to what Object is meant to convey. Some suggestions, based on description:

EdgeLifetime -> I'm not really opposed to this, although, you create EdgeLifetime objects in the code named 'connectors' which seems confusing to me.

EdgeTag -> Why not 'EdgeHandleTypeless', or 'EdgeHandleBase', something that conveys what its meant to be/do, and fits with most of our other naming schemes.

Off topic, but, Since its just holding an EdgeLifetime, is there a reason the edge destruction can't be handled in its destructor?

EdgeHandle -> Also seems fine.

EdgeHandleObj -> Really don't like the 'Obj' suffix, because it doesn't convey any information. Based on the description, something like 'EdgeDelegate' makes more sense to me. Because its not really a Handle, it contains a typeless handle+ some other information, and a callable interface.

EdgeTypePair -> Makes sense, describes exactly what it is.

EdgeHolder -> This seems fine too.

cpp/mrc/include/mrc/node/channel_holder.hpp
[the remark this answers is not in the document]
drobison00: I'm not entirely sure why KeyT needs to be a template parameter; but, what about something like:

```c++
template <typename KeyT=std::size_t, typename T>
class MultiEdgeHolder { ... }

template<typename T>
using EdgeHolder = MultiEdgeHolder<,T>
```

cpp/mrc/include/mrc/node/channel_holder.hpp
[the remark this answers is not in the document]
mdemoret-nv: Discussed in person. Resolving.

cpp/mrc/include/mrc/node/edge_adapter_registry.hpp
```diff
@@ -80,6 +87,8 @@ struct EdgeAdapterRegistry
     static std::map<std::type_index, source_adapter_fn_t> registered_source_adapters;
     static std::map<std::type_index, sink_adapter_fn_t> registered_sink_adapters;
 
+    static std::vector<ingress_adapter_fn_t> registered_ingress_adapters;
```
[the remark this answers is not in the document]
mdemoret-nv: Leaving it as is for now after discussions

python/mrc/_pymrc/include/pymrc/utilities/function_wrappers.hpp
mdemoret-nv: Leaving this for now

cpp/mrc/include/mrc/manifold/composite_manifold.hpp
mdemoret-nv: Went with Base since that was the naming scheme from before.

cpp/mrc/include/mrc/manifold/composite_manifold.hpp
```diff
@@ -39,6 +41,9 @@ class CompositeManifold : public Manifold
             .enqueue([this] {
                 m_ingress = std::make_unique<IngressT>();
                 m_egress  = std::make_unique<EgressT>();
+
+                // Then link them together
+                node::make_edge(*m_ingress, *m_egress);
```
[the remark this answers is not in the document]
mdemoret-nv: This just avoided the need for a third node to connect the ingress to the egress which existed before.

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1041 @@
...
+    {
+        m_is_armed = false;
+    }
+
+  private:
+    bool m_is_armed{true};
+    std::function<void()> m_fn;
+};
+
+class EdgeTag
```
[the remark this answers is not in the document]
mdemoret-nv: Discussed naming and adjusted. Resolving

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1041 @@
...
+    {
+        return m_is_connected;
+    }
+
+    void add_connector(std::function<void()>&& fn)
+    {
+        this->add_connector(EdgeLifetime(std::move(fn), true));
+    }
+
+    void add_connector(EdgeLifetime&& connector)
```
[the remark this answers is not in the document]
mdemoret-nv: Discussed naming in person. Resolving.

cpp/mrc/include/mrc/node/channel_holder.hpp
```diff
@@ -0,0 +1,1041 @@
...
+struct EdgeHandleObj
+{
+  public:
+    const EdgeTypePair& get_type() const
+    {
+        return m_type;
+    }
+
+  protected:
+    EdgeHandleObj(EdgeTypePair type_pair, std::shared_ptr<EdgeTag> edge_handle) :
```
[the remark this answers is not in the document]
mdemoret-nv: Discussed naming in person. Resolving.

cpp/mrc/include/mrc/node/channel_holder.hpp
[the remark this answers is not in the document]
mdemoret-nv: Key is a template parameter because you can have different types of separation between the output nodes. For example, broadcast just uses size_t, but some may use strings for named outputs, and we have a test case that uses an enum. Having it be a template parameter supports all these cases.

cpp/mrc/src/public/node/edge_builder.cpp
mdemoret-nv: Discussed in person. Resolving.

drobison00 (approved):
LGTM

PR #272, opened 2023-01-25
rename morpheus_utils_configure_boost_* to use new utilities naming

cwharris:
Required to fix a breaking change in https://github.com/nv-morpheus/utilities/pull/5

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -30,9 +30,7 @@ rapids_find_package(CUDAToolkit
 
 # Boost
 # =====
-# - Use static linking to avoid issues with system-wide installations of Boost.
-# - Use numa=on to ensure the numa component of fiber gets built
-morpheus_utils_configure_boost_boost_cmake()
+morpheus_utils_configure_boost()
 
 # UCX
 # ===


mdemoret-nv (approved):

mdemoret-nv:
I had to revert the change to configure boost due to this section in the old code: 

```cmake
  if (NOT Boost_ADDED)
    # Now add it to the list of packages to install
    rapids_export_package(INSTALL Boost
      ${PROJECT_NAME}-core-exports
      GLOBAL_TARGETS Boost::context Boost::fiber Boost::hana Boost::filesystem Boost::system
    )

    # Overwrite the default package contents
    file(WRITE "${CMAKE_BINARY_DIR}/rapids-cmake/${PROJECT_NAME}-core-exports/install/package_Boost.cmake" "find_dependency(Boost REQUIRED COMPONENTS context fiber filesystem system)")
  endif()
```
It's not immediately clear what this is doing, and its too risky to remove or blindly add back in at this point. Will need to readdress this after release.

mdemoret-nv:
Docs now broken. Reverting UCX

PR #277, opened 2023-01-27
Fix function wrappers with `functools.partial`

mdemoret-nv:
If you used `functools.partial` with any operator function, it would incorrectly thing you wanted to unpack tuples into arguments. This improves on the function signature handling to support using partial functions.

Also includes fixes to the conda build.


drobison00 (approved):
LGTM once all tests are passing.

PR #279, opened 2023-01-28
Add handling for empty downstream sources

mdemoret-nv:
Previously, we supported `source->node->[nothing]` where a source object didnt have any downstream sinks to write to. In that situation, the object was just discarded on writing. This fixes the edge functionality to replicate the previous system.


ryanolson (changes requested):
I feel like now is the time to enforce a null channel or a null cap. 

Nodes with dangling outputs should throw an error and not start. 

Nodes with dangling output can be null terminated which is the equivalent to dropping the data. 

If the Morpheus stages, if you want this behavior by default you can opinionated it there.

ryanolson (approved):

PR #286, opened 2023-02-03
[DRAFT] Mirror module / buffer + python bindings.

drobison00:


cpp/mrc/tests/modules/test_mirror_tap_util.cpp
```diff
@@ -0,0 +1,201 @@
...
+                                                           }
+
+                                                           sub.on_completed();
+                                                       });
+
+        auto sink = builder.make_sink<std::string>(test_name + "_main_sink", [&packets_main](std::string input) {
+            packets_main++;
+        });
+
+        // Untapped edge that will be broken and tapped by the mirror tap.
```
ryanolson, line 65: I'll make this comment here, but it applies broadly.

The comment suggests that you will take an edge and perform some surgery to insert your tap. This is a big PR, so I haven't really gotten down into the details yet, so the first question is, are you performing an edge break?

If so, can you explain the rational for why you choose this method over the prologue/epilogue taps that are currently in place?

For a given edge, you could either epilogue tap on the edge which would provide you access to the outbound data object before any channel effects (if transport is over a channel and not direct). Alternatively, you could prologue tap a sink and get the data after any channel policies were enforced (e.g. RecentChannel which might drop data).

I could have missed where you are using the epilogue/prologue taps as not all the file changes get populated via the web ui, but a quick search does not show any calls to `add_prologue_tap` or `add_epilogue_tap`.

ryanolson (dismissed):
Initial comments but we should follow up with a chat.

cpp/mrc/tests/modules/test_mirror_tap_util.cpp
drobison00: We can discuss, part of the approach is to improve easy of use/quality of life for end users, so that their initializer code doesn't actually need to be modified and this type of workflow can be supported:

```python
    tapped_init_wrapper_main = mirror_tap.tap(init_wrapper_main,
                                              test_name + "_main_source",
                                              test_name + "_main_sink")

    tapped_init_wrapper_mirrored = mirror_tap.stream_to(init_wrapper_mirrored,
                                                        test_name + "_mirror_sink")

    pipe.make_segment("segment_main", [], [mirror_tap.get_egress_tap_name()], tapped_init_wrapper_main)
    pipe.make_segment("segment_mirror", [mirror_tap.get_ingress_tap_name()], [], tapped_init_wrapper_mirrored)```

File: cpp/mrc/include/mrc/edge/edge_builder.hpp [hunk 2]
```diff
@@ -176,6 +176,89 @@ struct EdgeBuilder final
         sink.set_readable_edge_handle(edge);
     }
 
+    template <typename EdgeDataTypeT, typename SourceT, typename SinkT, typename SpliceInputT, typename SpliceOutputT>
+    static void make_edge_splice(SourceT& source, SinkT& sink, SpliceInputT& splice_input, SpliceOutputT& splice_output)
```
mdemoret-nv, line 180: What is the `splice_output` here? If you go from `source->sink` to `source->splice->sink`, how are there two parameters for splice?

cpp/mrc/include/mrc/edge/edge_builder.hpp
```diff
@@ -176,6 +176,89 @@ struct EdgeBuilder final
...
+            CHECK(splice_writable_provider != nullptr) << "Splice input is not a writable provider";
+
+            auto* splice_writable_acceptor = dynamic_cast<edge::IWritableAcceptor<EdgeDataTypeT>*>(&splice_output);
+            CHECK(splice_writable_acceptor != nullptr) << "Splice output is not a writable acceptor";
+
+            auto* writable_acceptor = dynamic_cast<edge::IWritableAcceptor<EdgeDataTypeT>*>(&source);
+            CHECK(writable_acceptor != nullptr) << "Source is not a writable acceptor";
+
+            auto& edge_holder = dynamic_cast<edge::EdgeHolder<EdgeDataTypeT>&>(*writable_acceptor);
+            CHECK(edge_holder.check_active_connection(false)) << "No active connection to splice into";
```
mdemoret-nv, line 219: Can the dynamic cast to `edge_holder` fail and return nullptr? This could have a null reference error on the next line.

cpp/mrc/include/mrc/modules/mirror_tap/mirror_tap_util.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include "mrc/version.hpp"
+
+#include <glog/logging.h>
+#include <nlohmann/json.hpp>
+
+#include <atomic>
+
+namespace mrc::modules {
+template <typename DataTypeT>
+class MirrorTapUtil
```
mdemoret-nv, line 38: I think of "Util" classes as static classes with a bunch of utility functions. I think `MirrorTapHelper` might be a better name for something you instantiate.

cpp/mrc/include/mrc/modules/properties/persistent.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+namespace mrc::modules {
+class PersistentModule
```
mdemoret-nv, line 21: Should this be removed?

cpp/mrc/include/mrc/modules/sample_modules.hpp
```diff
@@ -189,7 +189,7 @@ void TemplateModule<OutputTypeT>::initialize(segment::Builder& builder)
 template <typename OutputTypeT>
 std::string TemplateModule<OutputTypeT>::module_type_name() const
 {
-    return std::string(::mrc::type_name<type_t>());
+    return std::string(::mrc::boost_type_name<type_t>());
```
mdemoret-nv, line 192: Why the change from `mrc::type_name<type_t>` to `mrc::boost_type_name<type_t>`?

cpp/mrc/include/mrc/utils/type_utils.hpp
```diff
@@ -171,6 +172,12 @@ struct DataType
     TypeId m_type_id;
 };
 
+template <typename T>
+std::string boost_type_name()
+{
+    return boost::typeindex::type_id<T>().pretty_name();
+}
+
```
mdemoret-nv, line 180: How is this different than `type_name`?

cpp/mrc/src/public/segment/builder.cpp
```diff
@@ -41,6 +42,23 @@ std::string accum_merge(std::string lhs, std::string rhs)
...
+void Builder::init_module(std::shared_ptr<mrc::modules::SegmentModule> smodule)
+{
+    ns_push(smodule);
+    VLOG(2) << "Initializing module: " << m_namespace_prefix;
+    smodule->m_module_instance_registered_namespace = m_namespace_prefix;
+    smodule->initialize(*this);
+    ns_pop();
+
+    // TODO(Devin): Maybe a better way to do this with compile time type ledger.
+    if (std::dynamic_pointer_cast<modules::PersistentModule>(smodule) != nullptr)
```
mdemoret-nv, line 55: What is a `PersistentModule`? How is it different?

cpp/mrc/tests/modules/test_mirror_tap_stream.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <gtest/gtest-message.h>
+#include <gtest/gtest-test-part.h>
+
+#include <utility>
+#include <vector>
+
+using namespace mrc;
+
+// TODO(devin): Add tests
+TEST_F(TestMirrorTapStream, InitializationTest) {}
```
mdemoret-nv, line 36: Need some tests here

cpp/mrc/tests/test_edges.cpp
```diff
@@ -842,4 +842,41 @@ TEST_F(TestEdges, CreateAndDestroy)
...
+{
+    auto source = std::make_shared<node::TestSource<int>>();
+    auto node   = std::make_shared<node::TestNode<int>>();
+    auto sink   = std::make_shared<node::TestSink<int>>();
+
+    // Original edge
+    mrc::make_edge(*source, *sink);
+
+    // Tap edge
+    mrc::edge::EdgeBuilder::make_edge_splice<int>(*source, *sink, *node, *node);
```
mdemoret-nv, line 855: This seems odd to me. Why do we need to specify `*node` twice here? And shouldnt we be able to deduce the edge type if they are all the same `<int>`?

cpp/mrc/tests/test_edges.cpp
```diff
@@ -842,4 +842,41 @@ TEST_F(TestEdges, CreateAndDestroy)
...
+    // Original edge
+    mrc::make_edge(*source_rp, *sink_ra);
+
+    // Tap edge
+    mrc::edge::EdgeBuilder::make_edge_splice<int>(*source_rp, *sink_ra, *node, *node);
+
+    source->run();
+    node->run();
+    sink->run();
+}
```
mdemoret-nv, line 881: I feel like we need more tests here to ensure all of the combinations are covered

python/mrc/_pymrc/include/pymrc/segment.hpp
```diff
@@ -209,6 +209,14 @@ class BuilderProxy
                           std::shared_ptr<mrc::segment::ObjectProperties> source,
                           std::shared_ptr<mrc::segment::ObjectProperties> sink);
 
+    // TODO(devin) -- we can't pass all the variants, so we need to figure out how to do this
```
mdemoret-nv, line 212: Do we need to address this?

python/mrc/_pymrc/src/module_definitions/mirror_tap_util.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+    using python_mirror_tap_util_t = mrc::modules::MirrorTapUtil<pymrc::PyHolder>;
+    auto MirrorTap = py::class_<python_mirror_tap_util_t, std::shared_ptr<python_mirror_tap_util_t>>(smodule,
+                                                                                                     "MirrorTap");
+
+    MirrorTap.def(py::init(py::overload_cast<const std::string&>(&::MirrorTapUtilProxy::create)),
+                  py::return_value_policy::take_ownership);
+
+    MirrorTap.def(py::init(py::overload_cast<const std::string&, py::dict>(&::MirrorTapUtilProxy::create)),
+                  py::return_value_policy::take_ownership);
```
mdemoret-nv, line 99: As we discussed, I would prefer that the python module initialization code remain in the `python/mrc/core` where the modules will be created. All files under `python/mrc/_pymrc` are for creating shared utility classes used by the python modules.

python/mrc/_pymrc/src/module_definitions/segment_modules.cpp
```diff
@@ -0,0 +1,85 @@
...
+    SegmentModule.def("input_ids", &SegmentModuleProxy::input_ids);
+
+    SegmentModule.def("output_ids", &SegmentModuleProxy::output_ids);
+
+    // TODO(devin): need to think about if/how we want to expose type_ids to Python... It might allow for some nice
+    // flexibility
+    // SegmentModule.def("input_port_type_id", &SegmentModuleProxy::input_port_type_id, py::arg("input_id"))
+    // SegmentModule.def("input_port_type_ids", &SegmentModuleProxy::input_port_type_id)
+    // SegmentModule.def("output_port_type_id", &SegmentModuleProxy::output_port_type_id, py::arg("output_id"))
+    // SegmentModule.def("output_port_type_ids", &SegmentModuleProxy::output_port_type_id)
```
mdemoret-nv, line 83: Commented code

cpp/mrc/include/mrc/edge/edge_builder.hpp
mdemoret-nv, line 180: Why `make_edge_splice` instead of a more active name like `splice_edge`?

mdemoret-nv (changes requested):
A few changes that we discussed earlier about relocating files. Still looking into a few of the tap ring buffer implementations.

cpp/mrc/include/mrc/edge/edge_builder.hpp
drobison00: I actually like that terminology better, will update.

cpp/mrc/include/mrc/modules/mirror_tap/mirror_tap_util.hpp
drobison00: Agreed. Done.

cpp/mrc/include/mrc/modules/properties/persistent.hpp
drobison00: No, its used to test if a module needs to be persisted by the Builder. Usage is in `builder.cpp::init_module`, I'd looked at making it a variable on the module, but at the end of the day, its going to be something known at compile time and shouldn't be mutable.

cpp/mrc/include/mrc/modules/sample_modules.hpp
drobison00: mrc::type_name was a hand rolled approach; when I realized boost could do the same thing it felt reasonable to standardize on it as its likely to be more future proof.

cpp/mrc/include/mrc/edge/edge_builder.hpp
drobison00: **Also, the splice isn't always a single node, it can be an entire module, so we can't simplify by passing a single node and considering it the input and the output.

cpp/mrc/include/mrc/utils/type_utils.hpp
drobison00: The boost implementation is more robust in terms of parsing out the type name and error checking. The current `type_name` implementation is hand rolled from the early days of neo, when I wasn't aware that Boost could do it. I think it makes the module traces a little more readable.

cpp/mrc/src/public/segment/builder.cpp
drobison00: Its persistent/stateful. By default modules are stateless and can be released once they're initialized. A PersistentModule might have common member data; one motivating example is the stream buffer module that holds a boost circular buffer.

cpp/mrc/tests/test_edges.cpp
drobison00: Node is what is used to deduce if its an Acceptor or Provider, the EdgeBuilder API for splice edge doesn't assume that the SpliceInput and the SpliceOutput are the same node. One reason for this is that, one could be the input to a module, and one could be the output of the module, in which case they won't be the same node.

python/mrc/_pymrc/include/pymrc/segment.hpp
drobison00: For this, I was originally thinking about creating bindings that could take nodes and not just strings to look up nodes; at present, I don't think its worth the effort. There is going to be some additional work that needs to be done, at some point, to make creating splices using modules, from python, more user friendly. But its a fairly edge case issue, is will be non-breaking, and not required for this release.

cpp/mrc/tests/test_edges.cpp
drobison00: Which cases do you think are missing? The only paths we can't test here are string based lookups; all other invalid combinations will be compile time errors, so we can't write unit tests for them.

cpp/mrc/include/mrc/modules/properties/persistent.hpp
drobison00: ** Add comment indicating its intentionally empty.

cpp/mrc/tests/test_edges.cpp
drobison00: Added tests with each of the XXXComponent variants.

mdemoret-nv (approved):
Everything looks good now.

PR #288, opened 2023-02-08
Update CMake to only add fcoroutines flag if clang version is less than 16 + bump to latest utils

drobison00:
Clang will complain about deprecated flags if you pass fcoroutines to 16 or higher; this handles both cases and avoids warning spam.


dagardner-nv (changes requested):
Nevermind

dagardner-nv (approved):

PR #293, opened 2023-03-01
Pointer cast macro

dagardner-nv:
In debug it uses a dynamic cast and asserts not null, in release it uses a static cast.

In Morpheus I have a bunch of code that looks like:
```
    DCHECK(std::dynamic_pointer_cast<morpheus::InferenceMemory>(self.memory) != nullptr);
    return std::static_pointer_cast<morpheus::InferenceMemory>(self.memory);
```

This would allow for:
```
    return MRC_PTR_CAST(morpheus::InferenceMemory, self.memory);
```

Related to: https://github.com/nv-morpheus/Morpheus/issues/735

### DIFF
--- a/cpp/mrc/include/mrc/utils/macros.hpp
+++ b/cpp/mrc/include/mrc/utils/macros.hpp
@@ -29,7 +29,7 @@
     #define MRC_PTR_CAST(PTR_T, ptr) DCHECK_NOTNULL(std::dynamic_pointer_cast<PTR_T>(ptr))
 #endif
 
-#ifndef DELETE_COPYABILITYp
+#ifndef DELETE_COPYABILITY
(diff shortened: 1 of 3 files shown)


ryanolson (commented):
Could we simply make the macro be:

```

#define MRC_PTR_CAST(PTR_T, ptr)  \
    DCHECK_NOTNULL(std::dynamic_pointer_cast<PTR_T>(ptr)); \
    std::static_pointer_cast<PTR_T>(ptr)
#endif

I think there is a separate control mechanism for the `D` macros in glog separate from`NDEBUG`.

With this macro, we always do the dynamic check if glog's `DCHECK`s are enabled and skip them if not.

ryanolson (approved):

PR #299, opened 2023-03-13
Use gpu-latest for runner tags instead of specific driver version

jjacobelli:
Moving from using a specific driver version to the latest one. This will remove the need to always update the workflow every time a new driver is added to the CI runners

### DIFF
--- a/.github/workflows/ci_pipe.yml
+++ b/.github/workflows/ci_pipe.yml
@@ -132,7 +132,7 @@ jobs:
   test:
     name: Test
     needs: [build]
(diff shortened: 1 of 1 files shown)


dagardner-nv (changes requested):
We explicitly install `libnvidia-compute-520` in our `Dockerfile` to allow building on non-gpu hosts. So we would need to explicitly move to 525.

dagardner-nv (approved):

mdemoret-nv (approved):

PR #297, opened 2023-03-08
Devcontainer Updates

cwharris:
Changes similar to https://github.com/nv-morpheus/Morpheus/pull/717

PR #308, opened 2023-04-04
Upgrade to RMM 23.02

cwharris:
Adds RMM 23.02 support to MRC. Necessary for upgrading Morpheus to latest cudf. Also fixes a typo for the utilities repo rapids version, which was causing RMM conda-build to download and build with newer versions of RMM that was specified in the conda build matrix.

Based on #297


cwharris:
superceded by #309

PR #313, opened 2023-04-12
Dynamically loading `libnvidia-ml.so.1` instead of directly linking

mdemoret-nv:
Currently, we install the driver into one of the CI images to allow for stub generation during compilation. This was needed because `device_info.cpp` linked directly against `CUDA::nvml`. This caused a link dependency on a driver library which is problematic when building with CPU-only docker images.

Instead, we dynamically load `libnvidia-ml.so.1` (appending the `.so.1` to avoid collisions with the stub file `libnvidia-ml.so`) and the necessary functions at runtime. If the library is not found, using a GPU will be disabled. This allows loading of the library for stub generation without needing a GPU.

### DIFF
--- a/cpp/mrc/src/internal/system/device_info.cpp
+++ b/cpp/mrc/src/internal/system/device_info.cpp
@@ -60,6 +60,14 @@ struct NvmlState;
         throw std::runtime_error("Failed to load symbol " #function_var " from libnvidia-ml.so.1"); \
     }
 
+/**
+ * @brief This class wraps all calls to NVML and ensures that the library is dynamically loaded at runtime using
+ * `dlopen`. This is necessary to remove the link dependency to `libnvidia-ml.so` which is a driver library. Because
+ * driver libraries are loaded into the docker container via the NVIDIA container runtime, we do not want to directly
+ * link to them. As well, if the driver is not found, we can still run MRC without GPUs. This gives us an oportunity to
+ * catch the missing library instead of the OS throwing an error.
+ *
+ */
 struct NvmlHandle
 {
     NvmlHandle()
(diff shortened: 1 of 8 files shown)


dagardner-nv (approved):
LGTM, I wonder if there is any advantage to keeping container and test_container in the event that we ever do need additional packages for either the build or the test. 

Similarly I think we should update the `DOCKER_TARGET` array in `external/utilities/ci/runner/build_and_push.sh` script to :
```
DOCKER_TARGET=${DOCKER_TARGET:-"build" "test"}
```

Even if for MRC the build and test targets remain aliases for base.

mdemoret-nv:
> LGTM, I wonder if there is any advantage to keeping container and test_container in the event that we ever do need additional packages for either the build or the test.
> 
> Similarly I think we should update the `DOCKER_TARGET` array in `external/utilities/ci/runner/build_and_push.sh` script to :
> 
> ```
> DOCKER_TARGET=${DOCKER_TARGET:-"build" "test"}
> ```
> 
> Even if for MRC the build and test targets remain aliases for base.

Good point. Changed the names of the targets to `build` and `test` with the matching changes here: https://github.com/nv-morpheus/utilities/pull/30

cpp/mrc/src/internal/system/device_info.cpp
```diff
@@ -17,61 +17,163 @@
...
+
+/**
+ * @brief This class wraps all calls to NVML and ensures that the library is dynamically loaded at runtime using
+ * `dlopen`. This is necessary to remove the link dependency to `libnvidia-ml.so` which is a driver library. Because
+ * driver libraries are loaded into the docker container via the NVIDIA container runtime, we do not want to directly
+ * link to them. As well, if the driver is not found, we can still run MRC without GPUs. This gives us an oportunity to
+ * catch the missing library instead of the OS throwing an error.
+ *
+ */
+struct NvmlHandle
```
drobison00, line 71: Might make sense to have this be a singleton, or have a static counter to keep track of how many outstanding references to these symbols there are, which provides a way to check if anything is still using them when dlclose is called.

drobison00 (approved):

PR #317, opened 2023-04-13
Update to Python 3.10

cwharris:


python/setup.py
```diff
@@ -29,9 +29,8 @@
     classifiers=[
         "Intended Audience :: Developers",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
```
mdemoret-nv, line 32: This should be removed since we will only support 3.10

docs/quickstart/python/setup.cfg
```diff
@@ -23,7 +23,7 @@ classifiers =
     Intended Audience :: Developers
     Programming Language :: Python
     Programming Language :: Python :: 3.8
```
mdemoret-nv, line 25: Only support 3.10 so remove 3.8

docs/quickstart/hybrid/setup.cfg
```diff
@@ -23,7 +23,7 @@ classifiers =
     Intended Audience :: Developers
     Programming Language :: Python
     Programming Language :: Python :: 3.8
```
mdemoret-nv, line 25: Only support 3.10 so remove 3.8

ci/conda/recipes/libmrc/conda_build_config.yaml
```diff
@@ -26,8 +26,8 @@ cuda_compiler_version:
   - 11.8
 
 python:
-  - 3.8
   - 3.9
+  - 3.10
```
mdemoret-nv, line 30:
```suggestion
  - 3.10
```
We were optimistic that we would support multiple versions of python at first which would make this build matrix necessary. However, in `run_conda_build.sh` we just ignore this variant 100% of the time so we should just remove this.

ci/conda/recipes/run_conda_build.sh
```diff
@@ -99,9 +99,9 @@ fi
 # Choose default variants
 if hasArg quick; then
    # For quick build, just do most recent version of rapids
-   CONDA_ARGS_ARRAY+=("--variants" "{python: 3.8, rapids_version: 23.02}")
+   CONDA_ARGS_ARRAY+=("--variants" "{python: 3.10, rapids_version: 23.02}")
 else
-   CONDA_ARGS_ARRAY+=("--variants" "{python: 3.8}")
+   CONDA_ARGS_ARRAY+=("--variants" "{python: 3.10}")
```
mdemoret-nv, line 104: If we remove additional versions of python from the build matrix (`conda_build_config.yaml`), we can remove this flag which specifies that we only want to build with one variant.

mdemoret-nv (approved):
Looks good. A few small cleanup items. Before merging, please post a message in the group slack about the upgrade and any changes users will need to do to upgrade (i.e. `mamba env update ...`)

PR #320, opened 2023-04-20
Switch to using `cuda-toolkit` over `cudatoolkit`

mdemoret-nv:
## Description
This PR makes the following changes:

- Updates `dev_env.yml` to install `cuda-toolkit` over `cudatoolkit`
- We still have `cudatoolkit=11.8` listed at the end. This is necessary to specify which version to use. Some packages still use `cudatoolkit` and without specifying the version, 11.7 gets installed.
- Updated the conda build `meta.yml` to match `dev_env.yml`
- Also made some updates to remove warnings and hidden errors
- Updated the `utilities` repo to no longer require Cython unless its used
- Allowed us to remove Cython as a dependency in MRC

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [X] New or existing tests cover these changes.
- [X] The documentation is up to date with these changes.


drobison00 (commented):
It looks like Morpheus is having issues when we just install cuda-toolkit, vs cudatookit, on some internal dgx machines, either because CMake isn't reliably finding all the installed headers/libraries, or because they're not being installed correctly.

These changes should probably be tested in a completely clean environment to make sure they work outside of dev/ci.

dagardner-nv:
> It looks like Morpheus is having issues when we just install cuda-toolkit, vs cudatookit, on some internal dgx machines, either because CMake isn't reliably finding all the installed headers/libraries, or because they're not being installed correctly.

I think its the opposite, we're getting build issues when we install cudatookit, which are resolved by installing cuda-toolkit

drobison00:
> > It looks like Morpheus is having issues when we just install cuda-toolkit, vs cudatookit, on some internal dgx machines, either because CMake isn't reliably finding all the installed headers/libraries, or because they're not being installed correctly.
> 
> I think its the opposite, we're getting build issues when we install cudatookit, which are resolved by installing cuda-toolkit

My bad, I had it backwards :). Thanks David.

drobison00 (approved):
LGTM.

mdemoret-nv:
Merging to test with Morpheus. May need updates

PR #322, opened 2023-04-27
Update python debug detection for new version of `debugpy`

mdemoret-nv:
## Description
Looks like the library `debugpy` was updated to use cython instead of python. This broke our system of detecting when a debugger is attached.

This PR just changes the detection system to use `str(sys.gettrace())` instead of `sys.gettrace().__module__` since `__module__` is no longer set. The effect will still be the same.

## Checklist


cwharris:
What's with `init.sh` changing?

cwharris (approved):

dagardner-nv (approved):

drobison00 (approved):

mdemoret-nv:
> 

It was missing `chmod +x` in the file permissions so you got an error every time you tried to build the dev container.

PR #326, opened 2023-05-09
RxNodeComponent should set exceptions on the context

dagardner-nv:
## Description
This allows any thrown exceptions to be handled by the runner. 

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #327, opened 2023-05-10
Release an RxNodeComponent edge on error

dagardner-nv:
## Description
Followup to #326 

## Checklist

### DIFF
--- a/cpp/mrc/include/mrc/node/rx_node.hpp
+++ b/cpp/mrc/include/mrc/node/rx_node.hpp
@@ -235,6 +235,7 @@ class RxNodeComponent : public WritableProvider<InputT>, public WritableAcceptor
                 this->get_writable_edge()->await_write(std::move(message));
             },
             [this](std::exception_ptr ptr) {
+                WritableAcceptor<OutputT>::release_edge_connection();
                 runnable::Context::get_runtime_context().set_exception(std::move(ptr));
             },
             [this]() {
--- a/cpp/mrc/include/mrc/node/rx_node.hpp
+++ b/cpp/mrc/include/mrc/node/rx_node.hpp
@@ -235,9 +235,11 @@ class RxNodeComponent : public WritableProvider<InputT>, public WritableAcceptor
                 this->get_writable_edge()->await_write(std::move(message));
             },
             [this](std::exception_ptr ptr) {
+                runnable::Context::get_runtime_context().set_exception(std::move(ptr));
+            },
+            [this]() {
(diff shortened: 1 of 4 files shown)


cpp/mrc/include/mrc/node/rx_node.hpp
```diff
@@ -235,6 +235,7 @@ class RxNodeComponent : public WritableProvider<InputT>, public WritableAcceptor
                 this->get_writable_edge()->await_write(std::move(message));
             },
             [this](std::exception_ptr ptr) {
+                WritableAcceptor<OutputT>::release_edge_connection();
```
mdemoret-nv, line 238: Can you add a test to `test_edges.cpp` that checks for this? Should be pretty easy to throw an error in a node component. You will get an error if its not released correctly.

mdemoret-nv (approved):
Approved. But I think we should add tests to check for this on node component

cpp/mrc/include/mrc/node/rx_node.hpp
dagardner-nv: I can, but `test_segment.cpp` or `test_node.cpp` might be better locations for this. The reason is that in order to have a context set it needs to run inside of a segment, and there are currently no tests like this in `test_edges.cpp`

mdemoret-nv: lets put it in `test_node.cpp`

cpp/mrc/tests/test_edges.cpp
```diff
@@ -278,6 +279,31 @@ class TestNodeComponent : public NodeComponent<T, T>
...
+        return base_t::make_stream([this, fn](auto&&... args) {
+            m_stream_fn_called = true;
+            return fn(std::forward<decltype(args)>(args)...);
+        });
+    }
+
+    ~TestRxNodeComponent() override
+    {
+        // Debug print
+        VLOG(10) << "Destroying TestRxNodeComponent";
```
mdemoret-nv, line 301: Avoid debug logs in tests

cpp/mrc/tests/test_edges.cpp
```diff
@@ -278,6 +279,31 @@ class TestNodeComponent : public NodeComponent<T, T>
...
+    }
+
+    bool m_stream_fn_called = false;
```
mdemoret-nv, line 304: public members you can just drop the `m_`

mdemoret-nv (approved):

PR #333, opened 2023-06-20
use correct ostream.h header location

cwharris:


cwharris:
@GPUtester rerun tests

cwharris:
run tests

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ option(MRC_USE_CONDA "Enables finding dependencies via conda. All dependencies m
   environment" ON)
 option(MRC_USE_IWYU "Enable running include-what-you-use as part of the build process" OFF)
 
-set(MRC_RAPIDS_VERSION "23.02" CACHE STRING "Which version of RAPIDS to build for. Sets default versions for RAPIDS CMake and RMM.")
+set(MRC_RAPIDS_VERSION "23.04" CACHE STRING "Which version of RAPIDS to build for. Sets default versions for RAPIDS CMake and RMM.")
```
mdemoret-nv, line 39: Lets keep any rapids changes in a separate PR if necessary

mdemoret-nv (approved):

mdemoret-nv:
@cwharris Looks like `fmt/ostream.h` does not exist in our conda builds. Looks like we have 2 options:

1. Remove the include all together
   1. Since `fmt` comes with `spdlog`, it seems like this should work but you would need to test this in CI and locally for your NVAIE work
1. Add `fmt` as a dependency in conda
   1. Probably overkill, but this would ensure that the library can be found the same way for both conda and NVAIE

PR #355, opened 2023-07-13
Adding release procedures to Github Actions

mdemoret-nv:
## Description
Testing performing some release procedures in CI to improve release automation

### DIFF
--- a/.github/workflows/release_procedures.yml
+++ b/.github/workflows/release_procedures.yml
@@ -1,0 +1,81 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Release Procedures
+
+on:
+  workflow_dispatch:
+    inputs:
+      current_version:
+        description: 'Current version of the project'
+        required: true
+        type: string
+      next_version:
+        description: 'Next version of the project'
+        required: true
+        type: string
+      create_codefreeze_pr:
+        description: 'Creates/Updates the codefreeze PR'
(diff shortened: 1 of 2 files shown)


.github/workflows/release_procedures.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          lfs: false
+          path: 'mrc'
+          fetch-depth: 0
+      - name: Create PR
+        run: |
+          cat ci/release/pr_code_freeze_template.md | envsubst | \
+            gh pr create --base main --head branch-${VERSION} \
+              --title "[RELEASE] ${REPO_NAME} v${VERSION}" \
+              --body-file - \
```
cwharris, line 80: Remark: This is pretty cool. I didn't know you could pipe the file in like that.

.github/workflows/release_procedures.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          lfs: false
+          path: 'mrc'
+          fetch-depth: 0
+      - name: Create PR
+        run: |
+          cat ci/release/pr_code_freeze_template.md | envsubst | \
```
cwharris, line 77: Remark: Using envsubst here could be dangerous due `GITHUB_TOKEN` being in the environment. Wondering what the chances are we accidentally expose the token this way. I imagine it would require both updating the template _and_ running the workflow from the dashboard. Seems unlikely, but still possible.

Question: Is having `GITHUB_TOKEN` in the environment necessary?

Question: Are there other secrets that could be exposed with `envsubst`?

.github/workflows/release_procedures.yml
```diff
@@ -0,0 +1,81 @@
...
+    if: ${{ inputs.create_codefreeze_pr }}
+    name: Create Codefreeze PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          lfs: false
+          path: 'mrc'
+          fetch-depth: 0
```
cwharris, line 74: Question: Do we need to fetch all history for all branches and tags?

.github/workflows/release_procedures.yml
```diff
@@ -0,0 +1,81 @@
...
+      update_changelog:
+        description: 'Updates the CHANGELOG.md file for the current release and commits the changes'
+        required: true
+        type: boolean
+        default: false
+      merge_release_branch:
+        description: 'Merges the code freeze release branch, creates the release tag, and creates a new Github release'
+        required: true
+        type: boolean
+        default: false
```
cwharris, line 53: Remark: Looks like some of these inputs are unused.

Question: Should the unused inputs be removed?

cwharris (changes requested):

PR #359, opened 2023-07-28
Adopt pybind11 v2.10.4 and install via conda

dagardner-nv:
## Description
* Minor fix to `python/mrc/_pymrc/src/module_wrappers/shared_memory.cpp` when shmem is closed, the buffer will be none
* Update tests in `python/tests/test_edges.py` marked as `xfail` to `skip` since these are now triggering pybind asserts causing `SIGABRT`. This is expected as better error handling is listed as one of the new features.

## Checklist


python/tests/test_edges.py
```diff
@@ -371,7 +371,7 @@ def gen_parameters(*args,
 @pytest.mark.parametrize("source_type,sink_type",
                          [
                              pytest.param(m.Base, m.Base, id="source_base-sink_base"),
-                             pytest.param(m.Base, m.DerivedA, id="source_base-sink_derived", marks=pytest.mark.xfail),
+                             pytest.param(m.Base, m.DerivedA, id="source_base-sink_derived", marks=pytest.mark.skip),
```
cwharris, line 374: why switch to skip? does the test segfault?

cwharris (approved):
LGTM

python/tests/test_edges.py
dagardner-nv: Yes, it was already marked "xfail", but newer versions of pybind have better error checking and this test was triggering an assert

mdemoret-nv: These are marked `xfail` because they should generate a runtime error telling you that you cant make an edge between these types. The reason it was failing before was due to MRC throwing the desired exception, not due to an error in pybind. What is the new pybind11 assert that is getting triggered?

dagardner-nv:
We will likely need to continue installing pybind11 via CPM in the near future (#362). Will create a new PR moving to 2.10.4 via CPM.

PR #409, opened 2023-10-24
Move Pycoro from Morpheus to MRC

cwharris:
Moves pycoro from Morpheus to MRC and incorperates tests from https://github.com/nv-morpheus/Morpheus/pull/1286

Closes https://github.com/nv-morpheus/Morpheus/issues/1268

### DIFF
--- a/python/tests/test_pycoro.py
+++ b/python/tests/test_pycoro.py
@@ -14,12 +14,13 @@
 # limitations under the License.
 
 import asyncio
+
 import pytest
 
-from mrc.core import coro as pycoro # pylint: disable=morpheus-incorrect-lib-from-import
 from mrc._pymrc.tests.coro.coro import call_async
 from mrc._pymrc.tests.coro.coro import call_fib_async
 from mrc._pymrc.tests.coro.coro import raise_at_depth_async
+from mrc.core import coro as pycoro
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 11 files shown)


python/tests/test_pycoro.py
```diff
@@ -0,0 +1,152 @@
...
+        return ['a', 'b', 'c']
+
+    coros = [pycoro.wrap_coroutine(inner()) for _ in range(expected_count)]
+
+    returned_vals = await asyncio.gather(*coros)
+
+    end_time = asyncio.get_running_loop().time()
+
+    assert returned_vals == ['a'] * expected_count
+    # assert hit_count == expected_count
```
drobison00, line 72: Are these going to stay or be expanded on or can we remove?

drobison00 (approved):
Looks good overall; probably need to figure out some way to test the fiber based paths at some point, it doesn't look like anything hits them at the moment.

python/tests/test_pycoro.py
cwharris: @drobison00 yeah the only way to do that is with the CoroutineRunnable. That's a separate issue and the next PR I'm working on.

cwharris:
@drobison00 yeah the only way to do that is with the CoroutineRunnable. That's a separate issue and the next PR I'm working on.

python/tests/test_pycoro.py
cwharris: ah good catch, thanks. I'll add these back in. Had to comment them out while I was tracking down a bug in my test code.

PR #410, opened 2023-10-27
Add AsyncioRunnable

cwharris:
Moves the CoroutineRunnable from Morpheus' Sherlock feature branch to MRC and renames it to AsyncioRunnable as it is heavily dependent on asyncio. Adjustments were made such that the Scheduler would no longer own a task container and/or tasks, leaving the scheduler interface simpler. Instead, the runnable is responsible for the lifetime of the tasks it creates. This leaves the scheduler with a single responsibility.

Much of the code could be moved to MRC proper from PyMRC, but it's not immediately obvious where the code should live or whether it would be reused, so keeping it colocated with the AsyncioRunnable makes the most sense for now, imo.

### DIFF
--- a/ci/conda/environments/dev_env.yml
+++ b/ci/conda/environments/dev_env.yml
@@ -59,6 +59,7 @@ dependencies:
   - pybind11-stubgen=0.10
   - pytest
   - pytest-timeout
+  - pytest-asyncio
   - python=3.10
   - scikit-build>=0.17
   - sysroot_linux-64=2.17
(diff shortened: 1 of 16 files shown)


drobison00 (approved):
Approving with the assumption that asyncio_runnable.hpp and asyncio_scheduler.hpp are just copied over from Morpheus.

cwharris:
Closed because some of the changes were made under a faulty understanding of the asyncio event loop running mechanism. Fixed in https://github.com/nv-morpheus/MRC/pull/411

PR #411, opened 2023-10-30
Add AsyncioRunnable

cwharris:
Moves the CoroutineRunnable from Morpheus' Sherlock feature branch to MRC and renames it to AsyncioRunnable as it is heavily dependent on asyncio. Adjustments were made such that the Scheduler would no longer own a task container and/or tasks, leaving the scheduler interface simpler. Instead, the runnable is responsible for the lifetime of the tasks it creates. This leaves the scheduler with a single responsibility.

Much of the code could be moved to MRC proper from PyMRC, but it's not immediately obvious where the code should live or whether it would be reused, so keeping it colocated with the AsyncioRunnable makes the most sense for now, imo.


python/mrc/_pymrc/tests/test_asyncio_runnable.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+class MyAsyncioRunnable : public pymrc::AsyncioRunnable<int, unsigned int>
+{
+    mrc::coroutines::AsyncGenerator<unsigned int> on_data(int&& value) override
+    {
+        co_yield value;
+        co_yield value * 2;
+    };
+};
+
```
drobison00, line 61: I feel like we could do with a more robust set of tests, and more tests for each of the underlying components.

drobison00 (changes requested):

python/mrc/_pymrc/include/pymrc/asyncio_runnable.hpp
```diff
@@ -0,0 +1,398 @@
...
+#include <mrc/node/sink_properties.hpp>
+#include <mrc/runnable/forward.hpp>
+
+#include <coroutine>
+#include <exception>
+#include <functional>
+
+namespace mrc::pymrc {
+
+class ExceptionCatcher
```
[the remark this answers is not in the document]
cwharris: no reason other than convenience.

python/mrc/_pymrc/tests/test_asyncio_runnable.cpp
cwharris: I've added some more test cases, but the granularity is not great due to the nature of the MRC launch infrastructure. Generally happy path and exception paths are tested.

drobison00 (approved):

PR #437, opened 2024-02-05
Track remote descriptors

pentschev:
## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

### DIFF
--- a/cpp/mrc/CMakeLists.txt
+++ b/cpp/mrc/CMakeLists.txt
@@ -191,7 +191,7 @@ target_link_libraries(libmrc
     gRPC::grpc++
     gRPC::grpc
     gRPC::gpr
-    ucxx
+    ucxx::ucxx
   PRIVATE
     hwloc::hwloc
     prometheus-cpp::core # private in MR !199
--- a/cpp/mrc/CMakeLists.txt
+++ b/cpp/mrc/CMakeLists.txt
@@ -191,7 +191,7 @@ target_link_libraries(libmrc
     gRPC::grpc++
     gRPC::grpc
     gRPC::gpr
-    ucxx::ucxx
+    ucxx
   PRIVATE
     hwloc::hwloc
     prometheus-cpp::core # private in MR !199
--- a/cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
+++ b/cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
@@ -21,6 +21,7 @@
 #include "internal/data_plane/client.hpp"
 #include "internal/data_plane/server.hpp"
 #include "internal/memory/host_resources.hpp"
+#include "internal/remote_descriptor/messages.hpp"
 #include "internal/ucx/endpoint.hpp"
 #include "internal/ucx/ucx_resources.hpp"
 #include "internal/ucx/utils.hpp"
@@ -379,7 +380,40 @@ uint64_t DataPlaneResources2::register_remote_decriptor(
 {
     auto object_id = get_next_object_id();
     remote_descriptor->encoded_object().set_object_id(object_id);
+    m_remote_descriptor_by_id[object_id] = remote_descriptor;
     return object_id;
+}
+
+void DataPlaneResources2::decrement()
+{
+    std::map<std::shared_ptr<ucxx::Endpoint>, std::shared_ptr<ucxx::Request>> probeRequests;
+    for (auto& ep_pair : m_endpoints_by_address)
+    {
+        auto ep     = ep_pair.second;
+        auto worker = ep->getWorker();
+        if (worker->tagProbe(0))
+        {
+            remote_descriptor::RemoteDescriptorDecrementMessage dec_message;
+
+            auto decrement_request = ep->tagRecv(&dec_message,
+                                                 sizeof(remote_descriptor::RemoteDescriptorDecrementMessage),
+                                                 0);
+            while (!decrement_request->isCompleted())
+                worker->progress();
+
+            if (dec_message.tokens > 0)
+            {
+                auto remote_descriptor = m_remote_descriptor_by_id[dec_message.object_id];
+                auto tokens            = remote_descriptor->encoded_object().tokens();
+                tokens -= dec_message.tokens;
+                remote_descriptor->encoded_object().set_tokens(tokens);
+                if (tokens == 0)
+                {
+                    m_remote_descriptor_by_id.erase(dec_message.object_id);
+                }
+            }
+        }
+    }
 }
 
 // std::shared_ptr<ucxx::Request> DataPlaneResources2::receive_async2(void* addr,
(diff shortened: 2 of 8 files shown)


cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
```diff
@@ -363,6 +364,58 @@ std::shared_ptr<ucxx::Request> DataPlaneResources2::am_recv_async(std::shared_pt
...
 
+uint64_t DataPlaneResources2::get_next_object_id()
+{
+    auto object_id = m_next_object_id++;
+    while (m_remote_descriptor_by_id.find(object_id) != m_remote_descriptor_by_id.end())
+    {
+        object_id = m_next_object_id++;
+    }
+
+    return object_id;
```
mdemoret-nv, line 375: Can we just replace this with an atomic? We dont really care what the ID is, just that its unique.

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
```diff
@@ -363,6 +364,58 @@ std::shared_ptr<ucxx::Request> DataPlaneResources2::am_recv_async(std::shared_pt
...
+}
+
+void DataPlaneResources2::decrement()
+{
+    std::map<std::shared_ptr<ucxx::Endpoint>, std::shared_ptr<ucxx::Request>> probeRequests;
+    for (auto& ep_pair : m_endpoints_by_address)
+    {
+        auto ep     = ep_pair.second;
+        auto worker = ep->getWorker();
+        if (worker->tagProbe(0))
```
mdemoret-nv, line 394: Is it a requirement to receive have an endpoint for receiving a tagged message? If so, we may need to rethink this a bit to allow receiving messages from endpoints which have not been registered.

cpp/mrc/src/tests/test_network.cpp
```diff
@@ -84,12 +85,21 @@
 using namespace mrc;
 using namespace mrc::memory::literals;
 
+class DataPlaneResources2Tester : public data_plane::DataPlaneResources2
+{
+  public:
+    std::shared_ptr<runtime::RemoteDescriptorImpl2> get_descriptor(uint64_t object_id)
+    {
+        return m_remote_descriptor_by_id[object_id];
+    }
```
mdemoret-nv, line 94: We should discuss ways to avoid this pattern. Feel free to change visibility of any method as you see fit.

mdemoret-nv (commented):
Great progress. Lets discuss next steps to help expand the functionality

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
pentschev: Yes, using an atomic here makes sense. I'm not sure I understand your comment about uniqueness though, perhaps it's because the `find` which I added there to ensure we never reassign an ID that's in use (i.e., if we go over maximum number of IDs), which is probably unlikely given we use `uint64_t`, I can remove that check if you think we're reasonably confident we'll never go beyond the `uint64_t` limit.

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
pentschev: No, we don't need an endpoint here and we can use the worker instead. As I mentioned previously, I don't think using tags is a good solution and I just wrote it like this to have a working example. I have better idea in mind for this case, let's discuss it offline later today.

cpp/mrc/src/tests/test_network.cpp
pentschev: What would be the preferred solution here? Do you prefer that we make `m_remote_descriptor_by_id` public or something?

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
mdemoret-nv: I think its unlikely we will send more than 18,446,744,073,709,551,615 messages.

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
mdemoret-nv: Resolving since this is outdated

cpp/mrc/src/tests/test_network.cpp
mdemoret-nv: As we discussed, this is likely a fine solution.

cpp/mrc/CMakeLists.txt
```diff
@@ -191,7 +191,7 @@ target_link_libraries(libmrc
     gRPC::grpc++
     gRPC::grpc
     gRPC::gpr
-    ucxx::ucxx
+    ucxx
```
mdemoret-nv, line 194: Did we figure out why we needed `ucxx` here and not `ucxx::ucxx`? Can we switch this back to `ucxx::ucxx`?

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
```diff
@@ -160,6 +161,20 @@ DataPlaneResources2::DataPlaneResources2()
     DVLOG(10) << "initialize the registration cache for this context";
     m_registration_cache = std::make_shared<ucx::RegistrationCache2>(m_context);
 
+    auto decrement_callback = ucxx::AmReceiverCallbackType([this](std::shared_ptr<ucxx::Request> req) {
+        if (req->getStatus() != UCS_OK)
+            // TODO(Peter): Ensure the error gets raised somehow
+            DVLOG(10) << "Error calling decrement_callback";
```
mdemoret-nv, line 167: Our linter prefers brackets `{}` around all statements. Even one line for/if statements

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
```diff
@@ -363,6 +378,46 @@ std::shared_ptr<ucxx::Request> DataPlaneResources2::am_recv_async(std::shared_pt
     return request;
 }
 
+uint64_t DataPlaneResources2::get_next_object_id()
+{
+    // TODO(Peter): Handle overflow?
```
mdemoret-nv, line 383: We can ignore overflow. The max is 18,446,744,073,709,551,615 messages. At our current throughput of 11k/sec, that would take decades to hit that max value.

mdemoret-nv (approved):
Just a couple of quick comments and we can merge.

cpp/mrc/CMakeLists.txt
pentschev: Oh yes, good reminder `ucxx::ucxx` works now, I've reverted to the original via https://github.com/nv-morpheus/MRC/pull/437/commits/9d597bf0df8d3e52d7aa117c598727a42c627fd3 .

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
pentschev: Yeah, I wasn't sure how fast they could go. Removed comment via https://github.com/nv-morpheus/MRC/pull/437/commits/3e0cf32dfded9f96b9d539430f751d0bac47a1db .

PR #447, opened 2024-02-27
Allow memory pressure control in`DataPlaneResources2`

pentschev:
## Description
Allow memory pressure control in`DataPlaneResources2` via a maximum number of remote descriptors registered and add multithreaded test to verify completion despite blocking behavior of registration thread.

## Checklist

### DIFF
--- a/cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
+++ b/cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
@@ -391,7 +391,13 @@ uint64_t DataPlaneResources2::register_remote_decriptor(
 {
     auto object_id = get_next_object_id();
     remote_descriptor->encoded_object().set_object_id(object_id);
-    m_remote_descriptor_by_id[object_id] = remote_descriptor;
+    {
+        std::unique_lock lock(m_remote_descriptors_mutex);
+        m_remote_descriptors_cv.wait(lock, [this] {
+            return m_remote_descriptor_by_id.size() < m_max_remote_descriptors;
+        });
+        m_remote_descriptor_by_id[object_id] = remote_descriptor;
+    }
     return object_id;
 }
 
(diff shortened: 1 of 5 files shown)


cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
```diff
@@ -415,11 +421,20 @@ void DataPlaneResources2::decrement_tokens(remote_descriptor::RemoteDescriptorDe
...
+                m_remote_descriptor_by_id.erase(dec_message->object_id);
+            }
+            m_remote_descriptors_cv.notify_one();
         }
     }
 }
 
+void DataPlaneResources2::set_max_remote_descriptors(uint64_t max_remote_descriptors)
+{
+    m_max_remote_descriptors = max_remote_descriptors;
```
mdemoret-nv, line 435: If this can change while the tool is running, we probably need a `m_remote_descriptors_cv.notify_all();` to check any waiters.

cpp/mrc/src/internal/data_plane/data_plane_resources.hpp
```diff
@@ -219,6 +223,10 @@ class DataPlaneResources2
 
     void decrement_tokens(remote_descriptor::RemoteDescriptorDecrementMessage* dec_message);
 
+    uint64_t m_max_remote_descriptors{std::numeric_limits<uint64_t>::max()};
+    std::mutex m_remote_descriptors_mutex{};
+    std::condition_variable m_remote_descriptors_cv{};
```
mdemoret-nv, line 228: To be compatible with the rest of MRC, change these to use their `boost` counterparts: `boost::fibers::mutex` and `boost::fibers::condition_variable`. They should be functionally equivalent and work exactly the same if there is only one executing fiber.

mdemoret-nv (approved):
Looks good. Only two small changes and we can merge this.

cpp/mrc/src/internal/data_plane/data_plane_resources.cpp
pentschev: Makes sense, done in https://github.com/nv-morpheus/MRC/pull/447/commits/ba9be708fe523917954d0302155e95672c14c753, also add a test parameter to test for this condition as well in https://github.com/nv-morpheus/MRC/pull/447/commits/33aa01ca9bbe9c11b91bbe3b2e9db7771f596b01 .

cpp/mrc/src/internal/data_plane/data_plane_resources.hpp
pentschev: Ah yes, sorry for that, fixed in https://github.com/nv-morpheus/MRC/pull/447/commits/112790fdd9cfdb1ae08b9583feed0e6bba79ab1f .

mdemoret-nv (approved):

PR #451, opened 2024-03-11
Update cast_from_pyobject to throw on unsupported types rather than returning null

### LINKED ISSUE #450: [BUG]: mrc::pymrc::cast_from_pyobject should throw an exception rather than returning a None
### Version

24.03

### Which installation method(s) does this occur on?

_No response_

### Describe the bug.

Encountered this while working on nv-morpheus/Morpheus#1539 when `cast_from_pyobject` encounters an object which is not json serializable it returns a `None`.

Instead we should throw a `TypeError` since that is the error that the Python std `json.dumps` rasises.

### Minimum reproducible example

```shell
from morpheus.llm import LLMContext
ctx = LLMContext()
ctx.set_output(['a', RuntimeError('test'), 'b'])
print(ctx.view_outputs)
```

### Relevant log output

_No response_

### Full env printout

_No response_

### Other/Misc.

_No response_

### Code of Conduct

- [X] I agree to follow MRC's Code of Conduct
- [X] I have searched the [open bugs](https://github.com/nv-morpheus/MRC/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

dagardner-nv:
## Description
* Currently when `cast_from_pyobject` encounters an unsupported type it returns a json null.
* Updates the method to throw a `pybind11::type_error`, matching the `TypeError` exception raised by the Python std `json.dumps` method.
* Add `get_py_type_name` helper method
* Breaking behavior change

Closes #450 

## Checklist

### DIFF
--- a/python/mrc/_pymrc/src/utils.cpp
+++ b/python/mrc/_pymrc/src/utils.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/python/mrc/_pymrc/src/utils.cpp
+++ b/python/mrc/_pymrc/src/utils.cpp
@@ -76,37 +76,16 @@ const std::type_info* cpptype_info_from_object(py::object& obj)
     return nullptr;
 }
 
-std::string get_py_type_name(const pybind11::object& obj, bool ignore_exceptions)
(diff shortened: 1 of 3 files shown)


python/mrc/_pymrc/src/utils.cpp
```diff
@@ -170,8 +172,11 @@ json cast_from_pyobject(const py::object& source)
         return json(py::cast<std::string>(source));
     }
 
-    // else unsupported return null
-    return json();
+    // else unsupported return throw a type error
+    {
+        AcquireGIL gil;
+        throw py::type_error("Object is not JSON serializable");
+    }
```
mdemoret-nv, line 179: This error will be very frustrating for users in its current state. We need to report more information about the object that caused the exception. Especially since potentially deep in the object.

To this exception, we should add the typename of the object that caused the issue, a string representation of the object `py::str(source)` and the path of the object (you can get this with careful exception handlers and unwinding the recursive function).

mdemoret-nv (approved):

PR #455, opened 2024-03-13
Add JSONValues container for holding Python values as JSON objects if possible, and as pybind11::object otherwise

dagardner-nv:
## Description
* Add an optional `unserializable_handler_fn` callback to `mrc::pymrc::cast_from_pyobject` which will be invoked for any unsupported Python object. Allowing for serializing unsupported object.
ex:
```cpp
    pymrc::unserializable_handler_fn_t handler_fn = [](const py::object& source,
                                                       const std::string& path) {
        return nlohmann::json(py::cast<float>(source));
    };

    // decimal.Decimal is not serializable
    py::object Decimal = py::module_::import("decimal").attr("Decimal");
    py::object o       = Decimal("1.0");
    EXPECT_EQ(pymrc::cast_from_pyobject(o, handler_fn), nlohmann::json(1.0));
```

* Add `JSONValues` container class which is an immutable container for holding Python values as `nlohmann::json` objects if possible, and as `pybind11::object` otherwise. The container can be copied and moved, but the underlying `nlohmann::json` object is immutable.
* Updates `nlohmann_json` from 3.9 to 3.11 for `patch_inplace` method.
* Incorporates ideas from @drobison00's PR #417 with three key differences:

  1. Changes to the `cast_from_pyobject` are opt-in when `unserializable_handler_fn` is provided, otherwise there is no behavior change to the method.
  2. Unserializable objects are stored in `JSONValues` rather than a global cache.
  3. Does not rely on parsing the place-holder values.

This PR is related to nv-morpheus/Morpheus#1560

## Checklist

### DIFF
--- a/python/mrc/_pymrc/include/pymrc/utilities/json_values.hpp
+++ b/python/mrc/_pymrc/include/pymrc/utilities/json_values.hpp
@@ -17,7 +17,7 @@
 
 #pragma once
 
-#include "pymrc/types.hpp"
+#include "pymrc/types.hpp"  // for python_map_t & unserializable_handler_fn_t
 
 #include <nlohmann/json.hpp>
 #include <pybind11/pytypes.h>  // for PYBIND11_EXPORT & pybind11::object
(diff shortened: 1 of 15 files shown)


python/mrc/_pymrc/tests/test_json_values.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    std::vector<py::object> py_values = {py::cast(1), py::cast(2), py_dec, py::cast(4)};
+    py::list py_list                  = py::cast(py_values);
+
+    JSONValues j{py_list};
+    py::list result = j.to_python();
+    EXPECT_TRUE(result.equal(py_list));
+    py::object result_dec = result[2];
+    EXPECT_TRUE(result_dec.is(py_dec));
+}
```
drobison00, line 124: It may be worth checking the serialization code with a variety of python objects, like a module, a function, a generator,  etc.. and just make sure there aren't any unexpected gotchas.

python/mrc/_pymrc/include/pymrc/utilities/json_values.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    JSONValues set_value(const std::string& path, const pybind11::object& value) const;
+    JSONValues set_value(const std::string& path, nlohmann::json value) const;
+    JSONValues set_value(const std::string& path, const JSONValues& value) const;
+
+    std::size_t num_unserializable() const;
+    bool has_unserializable() const;
+
+    pybind11::object to_python() const;
+    nlohmann::json::const_reference to_json() const;
```
mdemoret-nv, line 60: 1. `view_json()` returns const ref to underlying nlohmann
2. `to_json(lambda)` returns copy of object and takes a handler for unserializable objects. Default handler throws
3. `JSONValues::stringify` static function where you could easily use it with `to_json(JSONValues::stringify)`

mdemoret-nv (changes requested):

PR #464, opened 2024-04-04
Add maximum simultaneous tasks support to `TaskContainer`

cwharris:
## Description
Adds maximum simultaneous tasks support to `TaskContainer` by only starting new tasks if the number of currently executing tasks is less than the maximum number of simultaneous tasks and starting new tasks as older tasks shut down. this eliminates the need for external semaphores or ticketing systems. 

## Checklist

### DIFF
--- a/cpp/mrc/include/mrc/coroutines/task_container.hpp
+++ b/cpp/mrc/include/mrc/coroutines/task_container.hpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/cpp/mrc/include/mrc/coroutines/task_container.hpp
+++ b/cpp/mrc/include/mrc/coroutines/task_container.hpp
@@ -139,6 +139,9 @@ class TaskContainer
      */
     auto gc_internal() -> std::size_t;
 
+    /**
+     * Starts the next taks in the queue.
+     */
     void start_next_task();
 
     /**
--- a/cpp/mrc/include/mrc/coroutines/task_container.hpp
+++ b/cpp/mrc/include/mrc/coroutines/task_container.hpp
@@ -60,7 +60,7 @@ class TaskContainer
      * @param e Tasks started in the container are scheduled onto this executor.  For tasks created
      *           from a coro::io_scheduler, this would usually be that coro::io_scheduler instance.
      */
-    TaskContainer(std::shared_ptr<Scheduler> e, std::size_t max_simultaneous_tasks = 0);
+    TaskContainer(std::shared_ptr<Scheduler> e, std::size_t max_concurrent_tasks = 0);
 
     TaskContainer(const TaskContainer&)                    = delete;
     TaskContainer(TaskContainer&&)                         = delete;
@@ -94,29 +94,19 @@ class TaskContainer
     auto garbage_collect() -> std::size_t;
 
     /**
-     * @return The number of tasks that are awaiting deletion.
-     */
-    auto delete_task_size() const -> std::size_t;
-
-    /**
-     * @return True if there are no tasks awaiting deletion.
-     */
-    auto delete_tasks_empty() const -> bool;
-
-    /**
      * @return The number of active tasks in the container.
      */
-    auto size() const -> std::size_t;
+    auto size() -> std::size_t;
 
     /**
      * @return True if there are no active tasks in the container.
      */
-    auto empty() const -> bool;
+    auto empty() -> bool;
 
     /**
      * @return The capacity of this task manager before it will need to grow in size.
      */
-    auto capacity() const -> std::size_t;
+    auto capacity() -> std::size_t;
 
     /**
      * Will continue to garbage collect and yield until all tasks are complete.  This method can be
@@ -141,7 +131,7 @@ class TaskContainer
     /**
      * Starts the next taks in the queue if one is available and max concurrent tasks has not yet been met.
      */
-    void try_start_next_task();
+    void try_start_next_task(std::unique_lock<std::mutex> lock);
 
     /**
      * Encapsulate the users tasks in a cleanup task which marks itself for deletion upon
@@ -161,7 +151,7 @@ class TaskContainer
     /// thread pools for indeterminate lifetime requests.
     std::mutex m_mutex{};
     /// The number of alive tasks.
-    std::atomic<std::size_t> m_size{};
+    std::size_t m_size{};
     /// Maintains the lifetime of the tasks until they are completed.
     std::list<std::optional<Task<void>>> m_tasks{};
     /// The set of tasks that have completed and need to be deleted.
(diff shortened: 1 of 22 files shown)


cpp/mrc/include/mrc/coroutines/task_container.hpp
```diff
@@ -166,6 +172,10 @@ class TaskContainer
     std::shared_ptr<Scheduler> m_scheduler_lifetime{nullptr};
     /// This is used internally since io_scheduler cannot pass itself in as a shared_ptr.
     Scheduler* m_scheduler{nullptr};
+    /// tasks to be processed in order of start
+    std::queue<decltype(m_tasks.end())> m_next_tasks;
+    /// maximum number of tasks to be run simultaneously
+    int32_t m_max_simultaneous_tasks;
```
mdemoret-nv, line 178: Why not have this as a `size_t` like the constructor signature and use 0 for unlimited number of tasks? A value of 0 would be invalid so we either need to check for that situation or just use 0 for the unlimited use case.

cpp/mrc/src/public/coroutines/task_container.cpp
```diff
@@ -53,20 +54,26 @@ auto TaskContainer::start(Task<void>&& user_task, GarbageCollectPolicy cleanup)
...
-    pos->value().resume();
+        // Store the task inside a cleanup task for self deletion.
+        auto pos  = m_tasks.emplace(m_tasks.end(), std::nullopt);
+        auto task = make_cleanup_task(std::move(user_task), pos);
+        *pos      = std::move(task);
+
+        m_next_tasks.push(pos);
+    }
+
+    if (m_max_simultaneous_tasks <= 0 or m_size <= m_max_simultaneous_tasks)
```
mdemoret-nv, line 73: `m_max_simultaneous_tasks` gets initialized by a `size_t` which is unsigned so it can never be negative.

cpp/mrc/src/public/coroutines/task_container.cpp
```diff
@@ -53,20 +54,26 @@ auto TaskContainer::start(Task<void>&& user_task, GarbageCollectPolicy cleanup)
 {
     m_size.fetch_add(1, std::memory_order::relaxed);
```
mdemoret-nv, line 55: Shouldnt `m_size` only be incremented once the task is launched? This will increment even when the task is queued

cpp/mrc/src/public/coroutines/task_container.cpp
```diff
@@ -133,6 +140,19 @@ auto TaskContainer::gc_internal() -> std::size_t
...
+{
+    auto pos = [this]() {
+        std::scoped_lock lk{m_mutex};
+        auto pos = m_next_tasks.front();
+        m_next_tasks.pop();
+        return pos;
+    }();
+
+    // Start executing from the cleanup task to schedule the user's task onto the thread pool.
+    pos->value().resume();
```
mdemoret-nv, line 153: I think this could be simplified by converting `m_mutex` to a `recursive_mutex`. Then you dont need to worry about deadlocks from the same thread.

cpp/mrc/src/public/coroutines/task_container.cpp
```diff
@@ -155,11 +175,19 @@ auto TaskContainer::make_cleanup_task(Task<void> user_task, task_position_t pos)
...
+        m_tasks_to_delete.push_back(pos);
+        // This has to be done within scope lock to make sure this coroutine task completes before the
+        // task container object destructs -- if it was waiting on .empty() to become true.
+        m_size.fetch_sub(1, std::memory_order::relaxed);
+    }
+
+    if (not m_next_tasks.empty())
+    {
+        start_next_task();
+    }
```
mdemoret-nv, line 189: Why is `m_next_tasks` not included in the scoped lock? Im concerned we might have a race condition. This class guards `m_next_tasks`, `m_tasks_to_delete` and `m_size` with a lock in other places of the code. However, we are reading the contents of `m_next_tasks` without the lock. Isnt is possible that a completing task could decrement `m_next_tasks` and trigger a race condition?

cpp/mrc/src/public/coroutines/task_container.cpp
mdemoret-nv, line 153: To be safe, we should add some sort of while check that can run "catch up". For example, if there is a weird unknown race condition which decrements `m_size` but does not launch a new task, you would end up with a situation where `m_size < (m_max_outstanding + 2)` and `m_next_tasks.size() == 2` (meaning 2 tasks could be started at once. A while loop could correct for that here.

mdemoret-nv (approved):
This is a cleaner solution than I was expecting. Have a couple suggestions on cleaning up the TaskContainer implementation to make it more resilient to race conditions.

cpp/mrc/include/mrc/coroutines/task_container.hpp
cwharris: Ah, I just forgot to change this to size_t. I think allowing any zero-or-negative number for unlimited tasks makes sense, otherwise we have to throw on negatives. size_t supports negatives.

cpp/mrc/src/public/coroutines/task_container.cpp
cwharris: m_size is the number of tasks in the container, not just the number of tasks currently executing.

cpp/mrc/include/mrc/coroutines/task_container.hpp
cwharris: I forgot to change this, and I always seem to forget that size_t is unsigned.

cpp/mrc/src/public/coroutines/task_container.cpp
cwharris: it's possible a different thread can execute the coroutine, so we need to make sure there's no lock at this point.

cwharris: the TaskContainer is the only thing that can start and stop these tasks, so if there is such a race condition, that would be a bug. catch up logic would just hide the bug.

cpp/mrc/src/public/coroutines/task_container.cpp
cwharris: totally possible. I'll fix it.

cpp/mrc/src/public/coroutines/task_container.cpp
cwharris: m_size is used to determine if there are any tasks left for cleanup and allow the final garbage collection coroutine to exit it's loop, so we need this to increment when a task is queued so we don't exit too early.

mdemoret-nv: Shouldnt we then have a separate variable for tracking the number of active tasks?

cwharris: I can't think of a reason to track the number of active tasks outside of making that information available for debugging purposes, and we don't have a use case for that yet. can you explain what that value would be used for?

mdemoret-nv (approved):
Looks better with the `try_start_next_task()` implementation. Just have the one question about `m_size`. Approved. Can merge when CI is passing. Just want to understand how it works with one variable.

cpp/mrc/src/public/coroutines/task_container.cpp
cwharris: After discussing, I realized my mistake in thinking that `m_size` was the right value to use to determine if we can start another task immediately. Instead, we can calculate the number of currently executing tasks by `m_size - m_next_tasks.size()`.

mdemoret-nv (approved):

mdemoret-nv (approved):

PR #496, opened 2024-08-30
Define a Python source which receives a reference to a subscriber

dagardner-nv:
## Description
* Allows a Python generator source to check if the subscriber is still subscribed.
* Define a class `SubscriberFuncWrapper`  for Python sources rather than just a lambda. The reason is that python objects captured by the lambda need to be destroyed while the gil is held, which causes a problem if the lambda is destroyed unexpectedly.
* Update `conftest.py` to set the loglevel to `DEBUG` if the `GLOG_v` environment variable is defined.

## Checklist

### DIFF
--- a/python/mrc/_pymrc/include/pymrc/segment.hpp
+++ b/python/mrc/_pymrc/include/pymrc/segment.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/python/mrc/_pymrc/include/pymrc/segment.hpp
+++ b/python/mrc/_pymrc/include/pymrc/segment.hpp
@@ -143,7 +143,7 @@ class BuilderProxy
                                                                        const std::string& name,
                                                                        pybind11::function gen_factory);
 
-    static std::shared_ptr<mrc::segment::ObjectProperties> make_subscriber_source(mrc::segment::IBuilder& self,
+    static std::shared_ptr<mrc::segment::ObjectProperties> make_source_subscriber(mrc::segment::IBuilder& self,
                                                                                   const std::string& name,
                                                                                   pybind11::function gen_factory);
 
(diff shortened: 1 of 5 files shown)


python/mrc/_pymrc/include/pymrc/segment.hpp
```diff
@@ -143,10 +143,14 @@ class BuilderProxy
...
-        const std::string& name,
-        const std::function<void(pymrc::PyObjectSubscriber& sub)>& f);
+    static std::shared_ptr<mrc::segment::ObjectProperties> make_subscriber_source(mrc::segment::IBuilder& self,
+                                                                                  const std::string& name,
+                                                                                  pybind11::function gen_factory);
+
+    // static std::shared_ptr<mrc::segment::ObjectProperties> make_source(
+    //     mrc::segment::IBuilder& self,
+    //     const std::string& name,
+    //     const std::function<void(pymrc::PyObjectSubscriber& sub)>& f);
```
mdemoret-nv, line 153: Excess comments

python/mrc/_pymrc/src/segment.cpp
```diff
@@ -257,6 +258,68 @@ std::shared_ptr<mrc::segment::ObjectProperties> build_source(mrc::segment::IBuil
...
+                subscriber.on_error(std::current_exception());
+                return;
+            }
+            subscriber.on_completed();
+
+            DVLOG(10) << ctx.info() << " Source complete";
+        };
+    }
+
+    py::function m_gen_factory{};
```
mdemoret-nv, line 320: We should use the `PyFuncWrapper` here to ensure we have the GIL when this is destroyed (then you dont need the destructor).

python/mrc/core/segment.cpp
```diff
@@ -134,6 +134,12 @@ PYBIND11_MODULE(segment, py_mod)
                                                                         const std::string&,
                                                                         py::function)>(&BuilderProxy::make_source));
 
+    Builder.def("make_subscriber_source",
```
mdemoret-nv, line 137: Probably would be better to call it `make_source_with_subscriber` or just `make_source_subscriber` so that `make_source` and `make_source_XXX` are near eachother in intellisense.

python/tests/test_executor.py
```diff
@@ -183,5 +249,35 @@ async def run_pipeline():
...
+
+    pipe = build_pipeline(source_fn, node_exception)
+
+    executor: mrc.Executor = None
+
+    executor = build_executor(pipe)
+
+    with pytest.raises(RuntimeError):
+        executor.join()
+
```
mdemoret-nv, line 280: Should this be in `test_node` instead of `test_executor`?

mdemoret-nv (approved):

python/tests/test_executor.py
dagardner-nv: It's in `test_executor.py` since there are other tests in this file which already test error handling of the executor.

PR #497, opened 2024-09-05
Change `LOG(WARNING)` to `VLOG(1)` when no GPUs are detected

dagardner-nv:
* Since CPU-only mode will become a supported feature we want to avoid unnecessary warnings.

Relates to nv-morpheus/Morpheus#1851

## Checklist


cpp/mrc/src/internal/system/device_info.cpp
```diff
@@ -150,7 +150,7 @@ struct NvmlState
             m_nvml_handle = std::make_unique<NvmlHandle>();
         } catch (std::runtime_error e)
         {
-            LOG(WARNING) << "NVML: " << e.what() << ". Setting DeviceCount to 0, CUDA will not be initialized";
+            DVLOG(1) << "NVML: " << e.what() << ". Setting DeviceCount to 0, CUDA will not be initialized";
```
mdemoret-nv, line 153: Lets just use `VLOG` instead of `DVLOG` here so we can still show it if needed

mdemoret-nv (approved):

PR #523, opened 2024-12-18
Fix bug where `AsyncioRunnable` hangs if `process_one` throws and the source is not emitting new values

dagardner-nv:
## Description
* Fixes a bug first observed in [NVIDIA-AI-Blueprints/vulnerability-analysis](https://github.com/NVIDIA-AI-Blueprints/vulnerability-analysis) and reported in nv-morpheus/Morpheus#2086
* `AsyncioRunnable` will now call `on_state_update(state_t::Kill)` when an exception is caught
* Replace blocking call to `await_read` with `await_read_until` allowing `AsyncioRunnable` to check `stop_source.stop_requested()` 
* Define new `await_read_until` method in `IEdgeReadable`, unfortunately this interface has numerous subclasses which all then needed new `await_read_until` methods, even though `EdgeChannelReader` is the only class that really needed it. Alternatives:
  - In `AsyncSink` perform a static cast of `this->get_readable_edge()` to `EdgeChannelReader`
  - Define `await_read_until` method in `IEdgeReadable` but give it an implementation that throws a non-impl exception (or asserts false)

## Checklist

### DIFF
--- a/cpp/mrc/include/mrc/node/sink_properties.hpp
+++ b/cpp/mrc/include/mrc/node/sink_properties.hpp
@@ -43,6 +43,13 @@ class NullReadableEdge : public edge::IEdgeReadable<T>
 
         return channel::Status::error;
     }
+
+    channel::Status await_read_until(T& t, const mrc::channel::time_point_t& tp) override
+    {
+        throw std::runtime_error("Attempting to read from a null edge. Ensure an edge was established for all sinks.");
+
+        return channel::Status::error;
+    }
 };
 
 /**
--- a/cpp/mrc/include/mrc/node/sink_properties.hpp
+++ b/cpp/mrc/include/mrc/node/sink_properties.hpp
@@ -40,15 +40,11 @@ class NullReadableEdge : public edge::IEdgeReadable<T>
     channel::Status await_read(T& t) override
     {
         throw std::runtime_error("Attempting to read from a null edge. Ensure an edge was established for all sinks.");
-
-        return channel::Status::error;
     }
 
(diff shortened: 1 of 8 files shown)


cpp/mrc/include/mrc/node/sink_properties.hpp
```diff
@@ -43,6 +43,13 @@ class NullReadableEdge : public edge::IEdgeReadable<T>
 
         return channel::Status::error;
     }
+
+    channel::Status await_read_until(T& t, const mrc::channel::time_point_t& tp) override
+    {
+        throw std::runtime_error("Attempting to read from a null edge. Ensure an edge was established for all sinks.");
```
willkill07, line 49: Do we want to throw or return an error status? Only one is strictly necessary.

In `source_properties.hpp` we only throw.

In `test_nodes.hpp` we do both.

willkill07 (changes requested):
One nit on consistency of unimplemented behavior. Otherwise LGTM.

cpp/mrc/include/mrc/node/sink_properties.hpp
dagardner-nv: In `sink_properties.hpp` this is for the `NullReadableEdge` class which shouldn't allow reads of any sort. I was following the implementation of `await_read` for the `await_read_until` method. In this case I think the `throw` is the correct thing. I wasn't sure why the `await_read` also has a `return channel::Status::error;` I suspect it was just there to satisfy a lint check.

willkill07 (approved):

PR #531, opened 2025-01-28
CI: Add new L4 runners for test

willkill07:
## Description
* Add two new CI runners on L4 (earliest and latest)
* Clean up the matrix definition

## Checklist
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.


.github/workflows/ci_pipe.yml
```diff
@@ -165,11 +165,12 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        arch: ["amd64", "arm64"]
+        arch: ["amd64"]
```
dagardner-nv, line 168: I don't think we want to remove `arm64`

willkill07: I didn't -- it's still in the include matrix below

willkill07: Did a force push maintaining the original style

PR #434, opened 2024-01-19
Revert partially constructed segments on-error in segment init function

dagardner-nv:
## Description
* Explicitly release edges from ingress/egress ports and manifolds.
* Ensure that the `inner_edge` held by the `ForwardingEgressProvider` is disconnected.
* Update `~EdgeHolder` to `LOG(FATAL)` in debug mode, `LOG(ERROR)` otherwise. The thinking here is that this error is only of use to MRC developers and not of use to downstream users.

Closes #360 

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/nv-morpheus/Morpheus/blob/main/docs/source/developer_guide/contributing.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #547, opened 2025-03-19
Add a pre-flight check when `pe_count > cpu_set.weight()`

dagardner-nv:
## Description
* Replace the `DCHECK_LE` in the `ReusableFiberEngineFactory` class with an if/throw.
* Update `~EdgeHolder` to `LOG(DFATAL)`. The thinking here is that this error is only of use to MRC developers and not of use to downstream users (Borrowed from PR #434).

## Checklist


cpp/mrc/include/mrc/edge/edge_holder.hpp
```diff
@@ -58,12 +59,37 @@ class EdgeHolder
...
+        bool is_connected = false;
+        if (m_connected_edge)
+        {
+            is_connected = m_connected_edge->is_connected();
+        }
+
+        ss << "\tis_connected=" << is_connected << "\tcheck_active_connection=" << this->check_active_connection(false);
+        return ss.str();
+    }
+
```
willkill07, line 92: Would no longer be necessary.
```suggestion
```

cpp/mrc/include/mrc/edge/edge_holder.hpp
```diff
@@ -58,12 +59,37 @@ class EdgeHolder
...
+            msg << "EdgeHolder(" << this << ") "
+                << "A node was destructed which still had dependent connections. Nodes must be kept alive while "
+                   "dependent connections are still active\n"
+                << this->connection_info();
+
+#if defined(NDEBUG)
+            LOG(ERROR) << msg.str();
+#else
+            LOG(FATAL) << msg.str();
+#endif
```
willkill07, line 72: `DFATAL` exists already, so we could just use that!

> The `DFATAL` severity logs a `FATAL` error in [debug mode](https://google.github.io/glog/0.7.1/logging/#debugging-support) (i.e., there is no `NDEBUG` macro defined), but avoids halting the program in production by automatically reducing the severity to `ERROR`.

```suggestion
            bool const is_connected = m_connected_edge && m_connected_edge->is_connected();
            LOG(DFATAL)
                << "EdgeHolder(" << this << ") "
                << "A node was destructed which still had dependent connections. Nodes must be kept alive while "
                   "dependent connections are still active\n"
                << "m_owned_edge=" << m_owned_edge.lock()
                << "\tm_owned_edge_lifetime=" << m_owned_edge_lifetime
                << "\tm_connected_edge=" << m_connected_edge
                << "\tis_connected=" << is_connected
                << "\tcheck_active_connection=" << this->check_active_connection(false);
```

cpp/mrc/include/mrc/edge/edge_holder.hpp
```diff
@@ -37,6 +37,7 @@
 #include <memory>
 #include <mutex>
 #include <optional>
+#include <sstream>  // for std::stringstream
```
willkill07, line 40:
```suggestion
```

willkill07 (changes requested):
Functionally looks great! Just a small nit on reusing `DFATAL` logging level to avoid stringstream + preprocessor branching.

cpp/mrc/include/mrc/edge/edge_holder.hpp
dagardner-nv: How is this not needed? if I am using `std::stringstream`?

dagardner-nv: Oh, I see now

willkill07 (approved):

